PR #5, opened 2024-01-23
Cluster-maximums workload

### LINKED ISSUE #4: [RFE] Cluster maximums OCP workload

Having a workload able to reproduce the documented cluster-maximums can be very useful to detect regressions of some components that are not that intensively used by the current workloads.

i.e.:

- Benchmarking max number of CRDs: It has been proven that a high number of CRDs had a negative impact in the API performance. Both in API responsiveness and resource usage. We're not tracking this scenario at the moment
- Max number of endpoints per service. In our current workloads, we're testing a high number of services, however we're not adding a high number of endpoints to them. This scenario is being currently tracked in upstream with kube-proxy implemented services, but we're not actually tracking it with OVNKubernetes

There more examples like the above. This new workload shouldn't be used as a rule of thumb to demonstrate the limits of a cluster, but as a new helper to detect and verify scenarios we're not currently tracking.

The cluster-maximums workload should be self-contained, based on a multi-job benchmark. With this approach maintaining and updating will be easier.

I started coding this workload, a initial approach about how it would look like is in the following snippet:

```yaml
# Would test 10k namespaces, 10k routes, 10k service, 20k pods and 30k network policies                                
  - name: max-namespaces                                                                                                                                      
    namespace: max-namespaces
    jobIterations: {{.NAMESPACES}}                                                                                     
    qps: {{.QPS}}                                     
    burst: {{.BURST}}                                                                                                  
    namespacedIterations: true                   
    waitWhenFinished: true                       
    preLoadImages: false                   # We don't need to preload since this job is reusing images previously used                                        
    jobPause: 2m                                                                                                                                              
    namespaceLabels:                   
      security.openshift.io/scc.podSecurityLabelSync: false                                                                                                   
      pod-security.kubernetes.io/enforce: privileged                                                                                                          
      pod-security.kubernetes.io/audit: privileged  
      pod-security.kubernetes.io/warn: privileged 
    objects:                                     
      - objectTemplate: deployment-server.yml              
        replicas: 1                                     
        inputVars: 
          podReplicas: 1                                                                                               
      - objectTemplate: deployment-client.yml
        replicas: 1                          
        inputVars:  
          podReplicas: 1                         
          ingressDomain: {{.INGRESS_DOMAIN}}            
      - objectTemplate: service.yml                        
        replicas: 1                                                                                                                     
      - objectTemplate: route.yml                                   
        replicas: 1                                                 
      - objectTemplate: np-deny-all.yml                             
        replicas: 1                                                 
      - objectTemplate: np-allow-from-clients.yml                   
        replicas: 1                                                 
      - objectTemplate: np-allow-from-ingress.yml                              
        replicas: 1                                                            
                                                                               
  - name: remove-max-namespaces                                                
    qps: 5                                                                     
    burst: 5                                                                   
    jobType: delete                                                            
    jobPause: 2m                                                               
    objects:                                                                   
      - kind: Namespace                                                                                                                                       
        labelSelector: {kube-burner-job: max-namespaces}                       

# 5k backends per service: Five times -> 5k server pods + 1 client pods + 1 route + 3 network policies                                                        
  - name: max-backends                                                         
    namespace: max-backends                                                    
    jobIterations: 5                                                           
    qps: {{.QPS}}                                                              
    burst: {{.BURST}}                                                          
    namespacedIterations: true                                                 
    waitWhenFinished: true                                                     
    preLoadImages: false             # We don't need to preload since this job is reusing images previously used                                              
    jobPause: 2m                                                               
    namespaceLabels:                                                           
      security.openshift.io/scc.podSecurityLabelSync: false                                                                                                   
      pod-security.kubernetes.io/enforce: privileged                           
      pod-security.kubernetes.io/audit: privileged                             
      pod-security.kubernetes.io/warn: privileged                              
    objects:                                                                   
      - objectTemplate: deployment-server.yml                                  
        replicas: 1                                                            
        inputVars:                                                             
          podReplicas: {{.BACKENDS}}                                           
      - objectTemplate: deployment-client.yml                                  
        replicas: 1                                                            
        inputVars:                                                             
          podReplicas: 1                                                       
          ingressDomain: {{.INGRESS_DOMAIN}}                                   
      - objectTemplate: service.yml                                            
        replicas: 1                                                            
      - objectTemplate: route.yml                                              
        replicas: 1                                                            
      - objectTemplate: np-deny-all.yml                                        
        replicas: 1                                                            
      - objectTemplate: np-allow-from-clients.yml                              
        replicas: 1                                                            
      - objectTemplate: np-allow-from-ingress.yml                              
        replicas: 1                                                            

  - name: remove-max-backends                                                  
    jobType: delete                                                            
    objects:                                                                   
      - kind: Namespace                                                        
        labelSelector: {kube-burner-job: max-backends}                      
```

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Implementing cluster-maximums workload

- Related Issue #
- Closes #4


cluster-maximums.go
```diff
@@ -0,0 +1,61 @@
...
+				log.Fatal("Error obtaining default ingress domain: ", err.Error())
+			}
+			os.Setenv("INGRESS_DOMAIN", ingressDomain)
+		},
+		Run: func(cmd *cobra.Command, args []string) {
+			wh.Run(cmd.Name(), getMetrics(cmd, "metrics-aggregated.yml"), alertsProfiles)
+
+		},
+	}
+	cmd.Flags().IntVar(&podsPerNode, "pods-per-node", 1000, "Pods per node")
```
jtaleric, line 55: 1000 pods per node?

cmd/config/cluster-maximums/cluster-maximums.yml
```diff
@@ -0,0 +1,141 @@
...
+        replicas: 1
+      - objectTemplate: route.yml
+        replicas: 1
+
+  - name: remove-max-pods-per-node
+    qps: 1
+    burst: 1
+    jobType: delete
+    jobPause: 2m
+    skipIndexing: true
```
jtaleric, line 47: Might be historical, but it might be interesting to capture the delete phase too wrt metric indexing.

cluster-maximums.go
rsevilla87: We support up to 2500, not sure if we should use that number...

cmd/config/cluster-maximums/cluster-maximums.yml
```diff
@@ -0,0 +1,141 @@
...
+        inputVars:
+          podReplicas: 2
+      - objectTemplate: deployment-client.yml
+        replicas: 1
+        inputVars:
+            podReplicas: 2
+            ingressDomain: {{.INGRESS_DOMAIN}}
+      - objectTemplate: service.yml
+        replicas: 1
+      - objectTemplate: route.yml
```
afcollins, line 39: Why are we creating a route for max pods per node test?

cmd/config/cluster-maximums/cluster-maximums.yml
```diff
@@ -0,0 +1,141 @@
...
+    defaultIndex: {{.ES_INDEX}}
+    type: {{.INDEXING_TYPE}}
+jobs:
+  - name: pods-per-node
+    namespace: pods-per-node
+    jobIterations: {{.JOB_ITERATIONS}}
+    qps: {{.QPS}}
+    burst: {{.BURST}}
+    namespacedIterations: true
+    iterationsPerNamespace: 50
```
afcollins, line 17: There are several dimensions here and I applaud you taking the initiative on codifying a max pods workload.

I am wondering whether this is the right value. It will result in namespaces capped at 50 iterations which breaks down into: 50x svcs, 50x server deployments, 100x server pods , 50x client deployments, 100x client pods.

If I were to propose this be a flag, then I think we would end up with flag explosion, so I may not make that proposal if we want to be opininoated.

I am curious whether we intend for these namespaces to be pseudo-realistic, or to target an actual maximum? (e.g. 6k services, which admittedly isn't an actual pod limitation).

cluster-maximums.go
afcollins: "support" is a scary word. :) 2500 is a maximum we tested at one time, but not a value we support as a blanket number for any cluster.

1000 pods could be a more reasonable, every-day value, but it also may not work if the cluster does not have enough capacity.

rsevilla87: Sure, that was one of the reasons I set it to 1K, 2.5K seemed a bit overkill to me to be tested on a more or less "regular" basis

cmd/config/cluster-maximums/cluster-maximums.yml
rsevilla87: I wanted to add some extra load to the data plane, not only stupid pods doing nothing

afcollins: Gotcha. I am not crazy about having the ingress overhead, but on the other hand, it would be a good test for the route->pod linkage for large numbers of pods.

vishnuchalla:
Implementation wise looks good to me. I hope all those defaults values are applicable to all the clusters where we plan to run this workload.

PR #46, opened 2024-04-04
Disable default cluster health check in BYOWL workloads

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Cluster health check shouldn't be executed in custom-workload and cluster-density by default. A user should opt-out this functionally with the flags.

- Related Issue #
- Closes #

### DIFF
--- a/cluster-density.go
+++ b/cluster-density.go
@@ -20,9 +20,7 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/kube-burner/kube-burner/pkg/config"
 	"github.com/kube-burner/kube-burner/pkg/workloads"
-	"github.com/openshift/client-go/config/clientset/versioned"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
 )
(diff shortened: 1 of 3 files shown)


cluster-density.go
```diff
@@ -38,15 +36,6 @@ func NewClusterDensity(wh *workloads.WorkloadHelper, variant string) *cobra.Comm
 		Use:   variant,
 		Short: fmt.Sprintf("Runs %v workload", variant),
 		PreRun: func(cmd *cobra.Command, args []string) {
-			kubeClientProvider := config.NewKubeClientProvider("", "")
```
vishnuchalla, line 41: Don't we want cluster health check in cluster-density?

rsevilla87: It can be enabled in all workloads https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/ocp.go#L85-L87
I removed it from cluster-density to standarize

vishnuchalla (approved):
lgtm

PR #49, opened 2024-04-15
Remove default metrics and alerts in BYOWL workloads

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

These workloads shouldn't require an alerts.yml and metrics.yml files in the same directory where they run, their metricsProfiles can be configured. At the moment it's only possible to configure metrics/alerts by using the metrics-endpoint flag, but once https://github.com/kube-burner/kube-burner/pull/611 gets merged, we should be able to configure them through the configuration file too

- Related Issue #
- Closes #


vishnuchalla (approved):
lgtm

PR #56, opened 2024-04-24
Making ocp-wrapper compatable with kube-burner v1.9.6

vishnuchalla:
## Type of change

- [x] Refactor
- [x] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Making code changes compatible with latest kb version .i.e. `v1.9.6`


josecastillolema:
From the web-burner workload perspective lgtm

cmd/ocp.go
```diff
@@ -89,6 +89,7 @@ func openShiftCmd() *cobra.Command {
 		kubeClientProvider := config.NewKubeClientProvider("", "")
 		wh = workloads.NewWorkloadHelper(workloadConfig, ocpConfig, kubeClientProvider)
 		envVars := map[string]string{
+			"ALERTS":        "alerts.yml",
```
rsevilla87, line 92: This variable should only be set when alerting is enabled

custom-workload.go
```diff
@@ -36,7 +36,10 @@ func CustomWorkload(wh *workloads.WorkloadHelper) *cobra.Command {
 				log.Fatalf("Error reading custom configuration file: %v", err.Error())
 			}
 			configFileName := strings.Split(configFile, ".")[0]
-			wh.Run(configFileName, nil, nil)
+			os.Unsetenv("ALERTS")
+			os.Unsetenv("METRICS")
+			os.Unsetenv("TS_METRICS")
```
rsevilla87, line 41: Why do you unset these, they are not set anywhere

vishnuchalla: Performing this check only for alerts just in case if alerting is enabled for custom-workload.

cmd/ocp.go
```diff
@@ -89,13 +89,17 @@ func openShiftCmd() *cobra.Command {
...
-			"GC_METRICS":    fmt.Sprintf("%v", gcMetrics),
-			"INDEXING_TYPE": indexer,
+			"ES_SERVER":      esServer,
+			"ES_INDEX":       esIndex,
+			"LOCAL_INDEXING": fmt.Sprintf("%v", localIndexing),
+			"QPS":            fmt.Sprintf("%d", QPS),
+			"BURST":          fmt.Sprintf("%d", burst),
+			"GC":             fmt.Sprintf("%v", gc),
+			"GC_METRICS":     fmt.Sprintf("%v", gcMetrics),
+			"INDEXING_TYPE":  indexer,
```
rsevilla87, line 99: I think we can remove references to INDEXING_TYPE, and the varible indexer, this variable was created to configure different indexers types in the same indexerConfig but since local and elasticsearch indexers are pre-configured (and not enabled) by default we don't need it

cmd/config/cluster-density-ms/cluster-density-ms.yml
```diff
@@ -8,11 +8,27 @@ global:
...
+      esServers: ["{{.ES_SERVER}}"]
+      insecureSkipVerify: true
+      defaultIndex: {{.ES_INDEX}}
+      type: {{.INDEXING_TYPE}}
+{{ end }}
+{{ if .LOCAL_INDEXING }}
+  - metrics: [{{.TS_METRICS}}]
+    indexer:
+      type: local
+      metricsDirectory: reporting-metrics
```
rsevilla87, line 25: Let's dump these metrics in the directory ts-metrics

cmd/config/cluster-density-ms/cluster-density-ms.yml
```diff
@@ -8,11 +8,27 @@ global:
...
+{{ if .ES_SERVER }}
+  - metrics: [{{.METRICS}}]
+    alerts: [{{.ALERTS}}]
+    indexer:
+      esServers: ["{{.ES_SERVER}}"]
+      insecureSkipVerify: true
+      defaultIndex: {{.ES_INDEX}}
+      type: {{.INDEXING_TYPE}}
+{{ end }}
+{{ if .LOCAL_INDEXING }}
```
rsevilla87, line 21:
```suggestion
{{ if eq .LOCAL_INDEXING "true" }}
```

cmd/ocp.go
```diff
@@ -89,13 +89,17 @@ func openShiftCmd() *cobra.Command {
...
-			"ES_SERVER":     esServer,
-			"ES_INDEX":      esIndex,
-			"QPS":           fmt.Sprintf("%d", QPS),
-			"BURST":         fmt.Sprintf("%d", burst),
-			"GC":            fmt.Sprintf("%v", gc),
-			"GC_METRICS":    fmt.Sprintf("%v", gcMetrics),
-			"INDEXING_TYPE": indexer,
+			"ES_SERVER":      esServer,
+			"ES_INDEX":       esIndex,
+			"LOCAL_INDEXING": fmt.Sprintf("%v", localIndexing),
```
rsevilla87, line 94:
```suggestion
			"LOCAL_INDEXING": fmt.Sprintf("%v", *localIndexing),
```

rsevilla87, line 94:
```suggestion
			"LOCAL_INDEXING": fmt.Sprintf("%v", *localIndexing),
```

cmd/config/cluster-density-ms/cluster-density-ms.yml
rsevilla87: We should check for the "true" string in the conditional

cmd/ocp.go
rsevilla87: We were setting a memory address in this env var LOL

cmd/config/cluster-density-ms/cluster-density-ms.yml
vishnuchalla: Sounds good to me. To only enable when applied.

cmd/ocp.go
vishnuchalla: Changed as suggested. Thanks

afcollins:
Thanks for doing this!

I believe with these changes, the [documentation](https://kube-burner.github.io/kube-burner-ocp/latest/#multiple-endpoints-support) is incorrect.

While we are touching that section, the command `oc sa new-token` is deprecated. When you run it, it says to use `oc create token` instead. We should also include that change as part of the docs change.

Also, by [hard-coding metricsDirectory](https://github.com/kube-burner/kube-burner-ocp/pull/56/files#diff-144cd869c5f7cc5180f3ff05855ca866a520baaefdb2a37abdddf1956efe5c3aR22) in the job config, we are reverting some functionality to generate the local dir with the uuid in the file name. We should preserve that functionality so we are not overwriting the same dir on repeated runs.

```
      metricsDirectory: reporting-metrics
```

afcollins:
I looked and the appending uuid seems to be removed upstream in kube-burner#611 . I guess I missed that there.

cmd/ocp.go
vishnuchalla: No, The existing snippet is correct. The `localIndexing` variable stores a pointer return by the command line flag that needs to be dereferenced in order to get its actual value. Another example [here](https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/ocp.go#L72).

vishnuchalla:
> Thanks for doing this!
> 
> I believe with these changes, the [documentation](https://kube-burner.github.io/kube-burner-ocp/latest/#multiple-endpoints-support) is incorrect.
> 
> While we are touching that section, the command `oc sa new-token` is deprecated. When you run it, it says to use `oc create token` instead. We should also include that change as part of the docs change.
> 
> Also, by [hard-coding metricsDirectory](https://github.com/kube-burner/kube-burner-ocp/pull/56/files#diff-144cd869c5f7cc5180f3ff05855ca866a520baaefdb2a37abdddf1956efe5c3aR22) in the job config, we are reverting some functionality to generate the local dir with the uuid in the file name. We should preserve that functionality so we are not overwriting the same dir on repeated runs.
> 
> ```
>       metricsDirectory: reporting-metrics
> ```

Thanks for all the inputs regarding docs @afcollins. Updated them accordingly.
Regarding metrics directory, I idea of this change is to leave it up to the user to provide a list of non-colliding metrics directory names for their runs in local. Let me know if you have any concerns around this idea.

cc: @rsevilla87

cmd/ocp.go
rsevilla87: I managed to run it w/o INDEXING_TYPE...

- No indexing
```shell
$ ./bin/amd64/kube-burner-ocp cluster-density-v2 --iterations 1 --churn=false
time="2024-04-26 10:09:11" level=info msg="🔥 Starting kube-burner (pull-56@7a62792ff2bd841618f4d74cb5703aec3fa8b821) with UUID 12e5f998-62c3-4dfa-820a-ca70ecd32b9b" file="job.go:99"
time="2024-04-26 10:09:11" level=info msg="📈 Registered measurement: podLatency" file="factory.go:91"                                                                                                                                        
...
time="2024-04-26 10:10:38" level=info msg="👋 Exiting kube-burner 12e5f998-62c3-4dfa-820a-ca70ecd32b9b" file="helpers.go:111"
```


- Local indexing

```shell
$ ./bin/amd64/kube-burner-ocp cluster-density-v2 --iterations 1 --churn=false --local-indexing                                                                                           
time="2024-04-26 10:15:57" level=info msg="📁 Creating local indexer: indexer-0" file="metrics.go:62"                                                                                                                                         
time="2024-04-26 10:15:57" level=info msg="👽 Initializing prometheus client with URL: https://prometheus-k8s-openshift-monitoring.apps.ocp-ci.perfscale.devcluster.openshift.com" file="prometheus.go:47"                                    
time="2024-04-26 10:15:58" level=info msg="📁 Creating local indexer: indexer-1" file="metrics.go:62"                                                                                                                                         
time="2024-04-26 10:15:58" level=info msg="👽 Initializing prometheus client with URL: https://prometheus-k8s-openshift-monitoring.apps.ocp-ci.perfscale.devcluster.openshift.com" file="prometheus.go:47"
time="2024-04-26 10:15:58" level=info msg="🔥 Starting kube-burner (pull-56@7a62792ff2bd841618f4d74cb5703aec3fa8b821) with UUID d0c3698e-2bda-426a-ac62-cc38b27d5db8" file="job.go:99"
time="2024-04-26 10:15:58" level=info msg="📈 Registered measurement: podLatency" file="factory.go:91"
...
time="2024-04-26 10:17:48" level=info msg="Indexing [1] documents from metric max-memory-workers" file="prometheus.go:225"                                                                                                                    
time="2024-04-26 10:17:48" level=info msg="File reporting-metrics/max-memory-workers.json created with 1 documents" file="prometheus.go:230"                                                                                                  
time="2024-04-26 10:17:48" level=info msg="Indexing [0] documents from metric max-memory-sum-infra-start" file="prometheus.go:225"                                                                                                            
time="2024-04-26 10:17:48" level=error msg="Empty document list in max-memory-sum-infra-start" file="prometheus.go:228"                                                                                                                       
time="2024-04-26 10:17:48" level=info msg="Indexing [0] documents from metric memory-cluster-usage-ratio-start" file="prometheus.go:225"                                                                                                      
time="2024-04-26 10:17:48" level=error msg="Empty document list in memory-cluster-usage-ratio-start" file="prometheus.go:228"                                                                                                                 
time="2024-04-26 10:17:48" level=info msg="Finished execution with UUID: d0c3698e-2bda-426a-ac62-cc38b27d5db8" file="job.go:276"                                                                                                              
time="2024-04-26 10:17:48" level=info msg="Garbage collecting jobs" file="job.go:289"                                                                                                                                                         
time="2024-04-26 10:17:48" level=info msg="Indexing cluster metadata document" file="helpers.go:143"                                                                                                                                          
time="2024-04-26 10:17:48" level=info msg="File reporting-metrics/clusterMetadata.json created with 1 documents" file="helpers.go:151"                                                                                                        
time="2024-04-26 10:17:48" level=info msg="Indexing cluster metadata document" file="helpers.go:143"                                                                                                                                          
time="2024-04-26 10:17:48" level=info msg="File reporting-metrics/clusterMetadata.json created with 1 documents" file="helpers.go:151"                                                                                                        
time="2024-04-26 10:17:48" level=info msg="👋 Exiting kube-burner d0c3698e-2bda-426a-ac62-cc38b27d5db8" file="helpers.go:111"                           
```


- os/es indexing
```
$ ./bin/amd64/kube-burner-ocp cluster-density-v2 --iterations=1 --churn=false --es-server ${ES_SERVER} --es-index kube-burner 
time="2024-04-26 11:49:28" level=info msg="📁 Creating opensearch indexer: indexer-0" file="metrics.go:62"
time="2024-04-26 11:49:30" level=info msg="👽 Initializing prometheus client with URL: https://prometheus-k8s-openshift-monitoring.apps.ocp-ci.perfscale.devcluster.openshift.com" file="prometheus.go:47"
time="2024-04-26 11:49:31" level=info msg="🔔 Initializing alert manager for prometheus: https://prometheus-k8s-openshift-monitoring.apps.ocp-ci.perfscale.devcluster.openshift.com" file="alert_manager.go:85"
time="2024-04-26 11:49:31" level=info msg="🔥 Starting kube-burner (pull-56@7a62792ff2bd841618f4d74cb5703aec3fa8b821) with UUID f69dc36f-a86c-4f25-a5bc-7fb5d3649318" file="job.go:99"
time="2024-04-26 11:49:31" level=info msg="📈 Registered measurement: podLatency" file="factory.go:91"
...
time="2024-04-26 11:53:46" level=info msg="Indexing [0] documents from metric cpu-infra-start" file="prometheus.go:225"
time="2024-04-26 11:53:46" level=info msg="Indexing skipped due to 0 docs" file="prometheus.go:230"
time="2024-04-26 11:53:46" level=info msg="Indexing [1] documents from metric cpu-workers" file="prometheus.go:225"
time="2024-04-26 11:53:47" level=info msg="Indexing finished in 442ms: created=1" file="prometheus.go:230"
time="2024-04-26 11:53:47" level=info msg="Finished execution with UUID: f69dc36f-a86c-4f25-a5bc-7fb5d3649318" file="job.go:276"
time="2024-04-26 11:53:47" level=info msg="Garbage collecting jobs" file="job.go:289"
time="2024-04-26 11:53:47" level=info msg="Indexing cluster metadata document" file="helpers.go:143"
time="2024-04-26 11:53:47" level=info msg="Indexing finished in 405ms: created=1" file="helpers.go:151"
time="2024-04-26 11:53:47" level=info msg="👋 Exiting kube-burner f69dc36f-a86c-4f25-a5bc-7fb5d3649318" file="helpers.go:111"
rsevilla@wonderland ~/labs/kube-burner-ocp (pull-56) $ echo $ES_SERVER 

``` 


I've done this changes only for the cd-v2 workload, and pushed a new commit on top of the code of this PR at https://github.com/kube-burner/kube-burner-ocp/commit/820e148178f73533c9dda185aa06d04d19059040

PTAL

rsevilla87:
> > Thanks for doing this!
> > I believe with these changes, the [documentation](https://kube-burner.github.io/kube-burner-ocp/latest/#multiple-endpoints-support) is incorrect.
> > While we are touching that section, the command `oc sa new-token` is deprecated. When you run it, it says to use `oc create token` instead. We should also include that change as part of the docs change.
> > Also, by [hard-coding metricsDirectory](https://github.com/kube-burner/kube-burner-ocp/pull/56/files#diff-144cd869c5f7cc5180f3ff05855ca866a520baaefdb2a37abdddf1956efe5c3aR22) in the job config, we are reverting some functionality to generate the local dir with the uuid in the file name. We should preserve that functionality so we are not overwriting the same dir on repeated runs.
> > ```
> >       metricsDirectory: reporting-metrics
> > ```
> 
> Thanks for all the inputs regarding docs @afcollins. Updated them accordingly. Regarding metrics directory, I idea of this change is to leave it up to the user to provide a list of non-colliding metric directory names for their runs in local. Let me know if you have any concerns around this idea.
> 
> cc: @rsevilla87

correct, that's the idea, but it doesn't work well though, all metrics get indexed to reporting-metrics, but that's due to a bug in the go-commons library... 

Regarding the issue mentioned by @afcollins, we could specify metricsDirectory with something like the below to fix it.

```
      metricsDirectory: reporting-metrics-{{.UUID}}
```

cmd/ocp.go
vishnuchalla: @rsevilla87 ohh, so do we want to hardcode `opensearch` as default indexer for all the config files?

vishnuchalla:
> > > Thanks for doing this!
> > > I believe with these changes, the [documentation](https://kube-burner.github.io/kube-burner-ocp/latest/#multiple-endpoints-support) is incorrect.
> > > While we are touching that section, the command `oc sa new-token` is deprecated. When you run it, it says to use `oc create token` instead. We should also include that change as part of the docs change.
> > > Also, by [hard-coding metricsDirectory](https://github.com/kube-burner/kube-burner-ocp/pull/56/files#diff-144cd869c5f7cc5180f3ff05855ca866a520baaefdb2a37abdddf1956efe5c3aR22) in the job config, we are reverting some functionality to generate the local dir with the uuid in the file name. We should preserve that functionality so we are not overwriting the same dir on repeated runs.
> > > ```
> > >       metricsDirectory: reporting-metrics
> > > ```
> > 
> > 
> > Thanks for all the inputs regarding docs @afcollins. Updated them accordingly. Regarding metrics directory, I idea of this change is to leave it up to the user to provide a list of non-colliding metric directory names for their runs in local. Let me know if you have any concerns around this idea.
> > cc: @rsevilla87
> 
> correct, that's the idea, but it doesn't work well though, all metrics get indexed to reporting-metrics, but that's due to a bug in the go-commons library...
> 
> Regarding the issue mentioned by @afcollins, we could conifgure metricsDirectory with something like the below to fix it.
> 
> ```
>       metricsDirectory: reporting-metrics-{{.UUID}}
> ```

Appending UUID at the end looks good to me, if we do not want any manual intervention at all.

cmd/ocp.go
rsevilla87: yeah, we're the main users of kube-burner-ocp, we can have some hardcoded stuff w/o issues :)

shashank-boyapally (approved):

cmd/config/custom-workload/custom-workload.yml
```diff
@@ -8,11 +8,19 @@ global:
...
+metricsEndpoints:
+{{ if .ES_SERVER }}
+  - indexer:
+      esServers: ["{{.ES_SERVER}}"]
+      insecureSkipVerify: true
+      defaultIndex: {{.ES_INDEX}}
+      type: opensearch
+{{ end }}
+{{ if eq .LOCAL_INDEXING "true" }}
+  - indexer:
```
afcollins, line 20: Is there a reason for custom-workload to be different from all other configs where only reporting metrics are indexed locally?

common.go
```diff
@@ -37,7 +39,8 @@ func getMetrics(cmd *cobra.Command, metricsProfile string) []string {
 	case Both:
 		metricsProfiles = []string{"metrics-report.yml", metricsProfile}
 	}
-	return metricsProfiles
+	os.Setenv("TS_METRICS", metricsProfile)
+	os.Setenv("METRICS", strings.Join(metricsProfiles, ","))
```
afcollins, line 43: I gather these are different vars for elastic and local indexing ..
I think these allow me to pick a different profile (e.g. aggregated vs non-aggregated) for elastic, but does not allow me to override for the local indexing.

Do we need separate vars here or can we reduce to the same `METRICS` ?

afcollins: > pick a different profile (e.g. aggregated vs non-aggregated)

Looking more, I see we don't get that CLI argument on other workloads, so I won't ask for that as it would be an enhancement.

Also, on the variable consolidation: I see how that allows there to be two different dirs locally for reporting and timeseries metrics .. I don't know that we need separate dirs TBH. I am fine writing to the same dir as it keeps things simpler and doesn't require this new var `TS_METRICS`

afcollins: Tested it locally and the split dir doesn't work anyway. Seems everything went to reporting dir.

```
$ $KUBE_BURNER_OCP version
Version: multiple-indexers
Git Commit: c1c8a2f2b5c4642e3ff48ace4aa7e9173b903794
Build Date: 2024-04-26-11:21:49
Go Version: go1.22.2
OS/Arch: darwin amd64

$ $KUBE_BURNER_OCP cluster-density-v2 --local-indexing --iterations=1
...
time="2024-04-26 11:38:37" level=info msg="Indexing cluster metadata document" file="helpers.go:143"
time="2024-04-26 11:38:37" level=info msg="File reporting-metrics-f85f60bf-8a5f-44bf-87bd-0e5841e2851d/clusterMetadata.json created with 1 documents" file="helpers.go:151"
time="2024-04-26 11:38:37" level=info msg="👋 Exiting kube-burner f85f60bf-8a5f-44bf-87bd-0e5841e2851d" file="helpers.go:111"

$ for i in  *f85f60bf* ; do echo "ls $i | wc" ; ls $i | wc ; done
ls reporting-metrics-f85f60bf-8a5f-44bf-87bd-0e5841e2851d | wc
     103     103    2696
ls ts-metrics-f85f60bf-8a5f-44bf-87bd-0e5841e2851d | wc
       0       0       0
```

afcollins (changes requested):
Thanks for all the updates!

I think the split local dir doesn't work as expected.

Unless we have some requirement for it, I would ask for it to be removed instead of looking into fixing it.  Personally, I am OK writing all metrics for a run to a single dir, and I think it simplifies the changes in this PR.

common.go
rsevilla87: Yes, there's a bug in the go-commons library, needs to be updated accordingly. We can address it in a future commit I don't consider this an stopper

afcollins: Sounds good to me

cmd/config/custom-workload/custom-workload.yml
vishnuchalla: Yes, because its just a code stub for reference we have decided to exclude timeseries and reporting metrics in one of our previous PRs: https://github.com/kube-burner/kube-burner-ocp/commit/7fe178eea60a59ba12e6c2105acade2ecd08d6af

common.go
vishnuchalla: To be honest its not a bug. Its more of an enhancement that we want to do at `go-commons` level in future. Currently it returns pointer references based on [indexer types](https://github.com/cloud-bulldozer/go-commons/blob/main/indexers/factory.go#L36) that we have defined in past. Meaning if we define multiple indexers with same type `local`, all of them will get overriden with the latest memory address.
```
time="2024-04-26 15:33:00" level=info msg="%!w(map[string]indexers.Indexer=map[indexer-0:0xc00029c9f0 indexer-1:0xc00029c9f0])" file="metrics.go:100"
```
That was the reason we were seeing all the metrics in a single dir. Anyways its non blocking and we can achieve the same with the updated configuration change in this PR. Let me know if you have any further thoughts/concerns.

rsevilla87:
Hi, I created a PR in go-commons to fix/improve the indexers code to allow creating multiple indexers of the same kind https://github.com/cloud-bulldozer/go-commons/pull/45

rsevilla87 (approved):
lgtm!

common.go
afcollins, line 43: Raising a point I think was lost in the other thread .. Do we need this separation of `TS_METRICS` and `METRICS` ? Can we not simply default to the same logic, which can be toggled via CLI arguments?

In all the configs and here, I would remove this split logic and reduce to `METRICS` only.

vishnuchalla: Yes, make sense. Without code actually splitting the indexing logic, we do not need a split in metrics. Keeping `METRICS` only as suggested.

afcollins: Thank you! :)

afcollins (approved):
Looks good to me!

Thank you for all the fixes for local indexing, Vishnu. :D

Issue #60, opened 2024-04-30
[BUG] web-burner-cluster-density workload is broken


josecastillolema:
# Bug Description

## **Output of `kube-burner-ocp` version**

```
$ kube-burner-ocp version
Version: 1.2.5
Git Commit: 37882f0e192f06f5008ad92d7bf0f55acb9c233c
Build Date: 2024-04-29T17:02:30Z
Go Version: go1.21.9
OS/Arch: linux amd64
```

## **Describe the bug**

There are some missing templates for the workload:
```
$ kube-burner-ocp web-burner-cluster-density --gc=false --sriov=true --alerting=
true --check-health=true --probe=false --bfd=false --limitcount=35 --scale 2 --crd=false --es-server=server --es
-index=index --profile-type reporting
time="2024-04-30 15:20:41" level=info msg="❤️ Checking for Cluster Health" file="cluster_health.go:45"
time="2024-04-30 15:20:41" level=info msg="Cluster is Healthy" file="cluster_health.go:53"
time="2024-04-30 15:20:41" level=info msg="📁 Creating opensearch indexer: indexer-0" file="metrics.go:62"
time="2024-04-30 15:20:42" level=info msg="👽 Initializing prometheus client with URL: https://prometheus-k
8s-openshift-monitoring.apps.vlan607.rdu2.scalelab.redhat.com" file="prometheus.go:47"
time="2024-04-30 15:20:42" level=info msg="🔔 Initializing alert manager for prometheus: https://prometheus
-k8s-openshift-monitoring.apps.vlan607.rdu2.scalelab.redhat.com" file="alert_manager.go:85"
time="2024-04-30 15:20:42" level=info msg="🔥 Starting kube-burner (1.2.5@37882f0e192f06f5008ad92d7bf0f55ac
b9c233c) with UUID 4423be6f-1f4b-4146-8dc9-cba58dafde78" file="job.go:99"
time="2024-04-30 15:20:42" level=info msg="📈 Registered measurement: podLatency" file="factory.go:91"
time="2024-04-30 15:20:43" level=info msg="Job cluster-density: 70 iterations with 30 ConfigMap replicas" f
ile="create.go:91"
time="2024-04-30 15:20:43" level=info msg="Job cluster-density: 70 iterations with 38 Secret replicas" file
="create.go:91"
time="2024-04-30 15:20:43" level=fatal msg="Error reading template node_density_pod_served.yml: open config
[...]

PR #61, opened 2024-04-30
Fix web-burner cluster-density workload

josecastillolema:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Closes https://github.com/kube-burner/kube-burner-ocp/issues/60

TODO: Add test cases for the web-burner-cluster-density scenario, today the CI only covers web-burner-node-density


cmd/config/web-burner-cluster-density/node_density_pod_served.yml
```diff
@@ -0,0 +1,49 @@
+---
```
vishnuchalla, line 1: Can this filename go with cluster_density* prefix to keep the naming consistent across the directory.

vishnuchalla:
@josecastillolema Looks like files in this PR got missed previously while adding web-burner cluster density workload. I think adding a CI test will prevent us from running into similar problems in future. So I would kindly request you to add a test in this PR itself, if it is not too much of work. Thanks.

josecastillolema:
@vishnuchalla all comments addressed ptal

josecastillolema:
@vishnuchalla kindly reminder, can we have this merged?
Thanks!

Issue #79, opened 2024-06-14
[BUG] cdv2 test failing to create build object on baremetal deployment


venkataanil:
# Bug Description
On OCP 4.14.19 barematal env, cluster-density-v2 is failing to create build object
`NAME                                            TYPE     FROM         STATUS                         STARTED   DURATION
build.build.openshift.io/cluster-density-v2-1   Docker   Dockerfile   New (InvalidOutputReference)    `

From our internal discussions 2 months ago, editing "configs.imageregistry.operator.openshift.io cluster" like below should fix the issue
storage:
    emptyDir: {}
    managementState: Managed

However this didn't completely fix the issue.

We have seen 2 builds in failed state
`[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-1
NAME                   TYPE     FROM         STATUS     STARTED          DURATION
cluster-density-v2-1   Docker   Dockerfile   Complete   24 minutes ago   18s
[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-2
NAME                   TYPE     FROM         STATUS                         STARTED   DURATION
cluster-density-v2-1   Docker   Dockerfile   New (InvalidOutputReference)             
[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-3
NAME                   TYPE     FROM         STATUS                         STARTED   DURATION
cluster-density-v2-1   Docker   Dockerfile   New (InvalidOutputReference)             
[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-4
NAME                   TYPE     FROM         STATUS     STARTED          DURATION
cluster-density-v2-1   Docker   Dockerfile   Complete   24 minutes ago   17s
[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-0
NAME                   TYPE     FROM         STATUS     STARTED          DURATION
cluster-density-v2-1   Docker   Dockerfile   Complete   24 minutes ago   18s`

I tried a hack of adding Iteration number to the build name i.e
`name: {{.JobName}}-{{.Iteration}}-{{.Replica}}`
 which worked for timebeing. I could see all build objects succesfully created.
`[root@e16-h12-b01-fc640 ~]# oc get build -n cluster-density-v2-1
NAME                     TYPE     FROM         STATUS     STARTED         DURATION
cluster-density-v2-1-1   Docker   Dockerfile   Complete   4 minutes ago   15s
[...]

PR #82, opened 2024-07-02
added image-registry check

shashank-boyapally:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Added image registry check for kube-burner-ocp, this can be used for bare-metal instances where image-registry co is not present by default.

- Related Issue #79
- Closes #

### DIFF
--- a/cluster-density.go
+++ b/cluster-density.go
@@ -20,6 +20,7 @@ import (
 	"strconv"
 	"time"
 
+	"github.com/kube-burner/kube-burner/pkg/config"
 	"github.com/kube-burner/kube-burner/pkg/workloads"
 	log "github.com/sirupsen/logrus"
 	"github.com/spf13/cobra"
@@ -52,6 +53,11 @@ func NewClusterDensity(wh *workloads.WorkloadHelper, variant string) *cobra.Comm
 			os.Setenv("INGRESS_DOMAIN", ingressDomain)
 		},
 		Run: func(cmd *cobra.Command, args []string) {
+			kubeClientProvider := config.NewKubeClientProvider("", "")
+			clientset, _ := kubeClientProvider.ClientSet(0, 0)
+			if !clusterImageRegistryCheck(clientset) {
+				log.Errorf("image-registry deployment is not deployed")
+			}
 			setMetrics(cmd, "metrics-aggregated.yml")
 			wh.Run(cmd.Name())
 		},
--- a/cluster-density.go
+++ b/cluster-density.go
@@ -53,10 +53,12 @@ func NewClusterDensity(wh *workloads.WorkloadHelper, variant string) *cobra.Comm
 			os.Setenv("INGRESS_DOMAIN", ingressDomain)
 		},
 		Run: func(cmd *cobra.Command, args []string) {
-			kubeClientProvider := config.NewKubeClientProvider("", "")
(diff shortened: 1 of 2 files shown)


rsevilla87:
Isn't the registry image registry status already verified by the cluster operators check?

https://github.com/kube-burner/kube-burner-ocp/blob/4a36d42712bd22e9308035149a3de93b33896f54/cluster_health.go#L67-L75

🤔

rsevilla87:
I think this PR won't be necessary, IUC, Anil wasn't passing the `--check-health`  flag to the kube-burner-ocp command

rsevilla87:
@shashank-boyapally , maybe we should default to true the value of this flag?, wdyt @jtaleric @afcollins @venkataanil ?

shashank-boyapally:
I was thinking since we get the list of installed `operators`, we cannot check the condition of `image-registry` if it is not present, hence I added an additional check, if I'm not mistaken image-registry is not present when we spin up a bm instance, please correct me if I'm wrong. Yes, we can enable the flag to default true.

shashank-boyapally:
Looks like the image-registry is present, hence we might not need explicit check to image-registry. Please confirm if I can close this PR as not needed anymore.

venkataanil:
Its better we get a BM deployment and reporduce this issue and then find root cause and fix it.

afcollins:
> find root cause
@venkataanil We know it already. Bare metal deployments do not include the image-registry by default.

It is [documented](https://docs.openshift.com/container-platform/4.16/registry/configuring_registry_storage/configuring-registry-storage-baremetal.html#registry-removed_configuring-registry-storage-baremetal) as a Day 2 operation.

Configuring a registry is not something kube-burner would do.

cluster-density.go
```diff
@@ -52,6 +53,11 @@ func NewClusterDensity(wh *workloads.WorkloadHelper, variant string) *cobra.Comm
 			os.Setenv("INGRESS_DOMAIN", ingressDomain)
 		},
 		Run: func(cmd *cobra.Command, args []string) {
+			kubeClientProvider := config.NewKubeClientProvider("", "")
```
rsevilla87, line 56: Let's use this check only for the `cluster-density-v2` variant

cluster_health.go
```diff
@@ -92,3 +92,16 @@ func isClusterHealthy(clientset kubernetes.Interface, openshiftClientset *versio
...
 }
+
+func clusterImageRegistryCheck(clientset kubernetes.Interface) bool {
+	deployment, err := clientset.AppsV1().Deployments("openshift-image-registry").Get(context.TODO(), "image-registry", metav1.GetOptions{})
+	if err != nil {
+		log.Errorf("Error getting deployment: %v", err)
+		return false
+	}
+	if deployment.Status.AvailableReplicas > 0 {
+		log.Infof("Deployment image-regsitry in namespace openshift-image-registry is available with %d replicas", deployment.Status.AvailableReplicas)
```
afcollins, line 103: Is this an Info or Debug level message?

shashank-boyapally: I can move it to Debug level if we consider as debug..

afcollins: I think it is at most a debug

rsevilla87 (approved):
lgtm

PR #89, opened 2024-07-19
Don't set indexing vars when metrics-endpoint is set

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

When using metrics-endpoint we shouldn't set any environment variables used for indexing

- Related Issue #
- Closes #

### DIFF
--- a/cmd/ocp.go
+++ b/cmd/ocp.go
@@ -81,17 +81,22 @@ func openShiftCmd() *cobra.Command {
 		wh = workloads.NewWorkloadHelper(workloadConfig, ocpConfig, kubeClientProvider)
 		wh.MetricsMetadata = make(map[string]interface{})
 		envVars := map[string]string{
-			"UUID":           workloadConfig.UUID,
-			"ES_SERVER":      esServer,
-			"ES_INDEX":       esIndex,
-			"LOCAL_INDEXING": fmt.Sprintf("%v", *localIndexing),
-			"QPS":            fmt.Sprintf("%d", QPS),
-			"BURST":          fmt.Sprintf("%d", burst),
-			"GC":             fmt.Sprintf("%v", gc),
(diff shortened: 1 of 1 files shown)


cmd/ocp.go
```diff
@@ -80,17 +80,22 @@ func openShiftCmd() *cobra.Command {
...
+			"GC":         fmt.Sprintf("%v", gc),
+			"GC_METRICS": fmt.Sprintf("%v", gcMetrics),
 		}
+		envVars["LOCAL_INDEXING"] = fmt.Sprintf("%v", *localIndexing)
 		if alerting {
 			envVars["ALERTS"] = "alerts.yml"
+		} else {
+			envVars["ALERTS"] = ""
+		}
+		// If metricsEndpoint is not set, use values from flags
```
vishnuchalla, line 95: Are we making sure that we are giving higher precedence to `metricEndpoint` over ES CLI args in the kube-burner core repo too?

rsevilla87: There're not ES CLI args in kube-burner core, they only exist in `kube-burner-ocp` (for now)

vishnuchalla: Looks like there is one: https://github.com/kube-burner/kube-burner/blob/main/cmd/kube-burner/kube-burner.go#L356. Which can be removed I guess. wdyt?

vishnuchalla: Yes it doesn't. But when we have metric endpoint do we want to keep es-server as a CLI arg?

rsevilla87: Not sure if Im following you... as I said before, the ES_SERVER and ES_INDEX env vars are not used by kube-burner core. I've added a small hint in the `--metrics-endpoint` help to note this behavior

vishnuchalla (approved):
As discussed offline, approving these changes.

PR #100, opened 2024-08-30
Fix golangci-lint precommit hook

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

The previous hook wasn't working as expected as it was relying on this configuration https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml#L1-L8 `golangci-lint run --new-from-rev HEAD --fix`, the hook's command basically fixes linting errors when the current code is different from HEAD (so no action is taken when changes are committed), this is not what we actually need, we want to report and hang on linting errors caused by a PR.

Also fixing some pieces of code to meet linting.

- Related Issue #
- Closes #

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -14,7 +14,6 @@ linters:
     - goimports
     - dupl
     - unparam
-    - revive
     - staticcheck
     - gosimple
     - unconvert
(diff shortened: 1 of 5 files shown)


vishnuchalla (approved):
lgtm

PR #101, opened 2024-09-03
Code changes to capture node bootup times for self-managed cluster

vishnuchalla:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Workers scale draft PR to describe the idea. Still in progress and changes will be continued.
More context: [Capturing node bootup times in AWS](https://docs.google.com/document/d/196igUG40JdrzYTzsbo5pCk4Wgx9hm4hziyHV4xHkye8/edit#heading=h.9pnu2zgzz0ij)
Pre-requisite: https://github.com/kube-burner/kube-burner/pull/687

- Closes # https://issues.redhat.com/browse/PERFSCALE-3364

## Testing
Tested in local with multiple scenarios
* Manual scaling: https://gist.github.com/vishnuchalla/3e7bab0105a20e2404ea6940a9e92655
* Auto scaling: https://gist.github.com/vishnuchalla/c5f3c7f36af2033b418280b0fb9ae40b
* Already scaled cluster: https://gist.github.com/vishnuchalla/0294b86ffbac77c648e4c42d317157e1


go.mod
```diff
@@ -8,15 +8,19 @@ require (
...
+	github.com/openshift/client-go v0.0.0-20240821135114-75c118605d5f
 	github.com/praserx/ipconv v1.2.1
 	github.com/sirupsen/logrus v1.9.3
 	github.com/spf13/cobra v1.8.0
+	k8s.io/api v0.30.1
 	k8s.io/apimachinery v0.30.1
 	k8s.io/client-go v0.30.1
 )
 
+replace github.com/kube-burner/kube-burner => /home/vchalla/myforks/kube-burner
```
vishnuchalla, line 21: Will be replaced accordingly after the pre-requisite PR is merged.

pkg/workers_scale/autoscaler.go
```diff
@@ -0,0 +1,292 @@
...
+)
+
+type AutoScalerScenario struct {}
+
+// Returns a new scenario object
+func (awsAutoScalerScenario *AutoScalerScenario) OrchestrateWorkload(scaleConfig ScaleConfig) {
+	var err error
+	kubeClientProvider := config.NewKubeClientProvider("", "")
+	clientSet, restConfig := kubeClientProvider.ClientSet(0, 0)
+	nodeCount, _ := getNodeCount(clientSet)
```
rsevilla87, line 45: Node count is already provided by metadata

vishnuchalla: Removed it as it is not necessary.

rsevilla87 (approved):
As we already discussed in slack, this workload doesn't make an extensive usage of the kube-burner functionalities , it's more similar to a script than an actual workload. Most of actions are executed in a not `kube-burner` way, as the scripted actions of the logic already take care of the lifecycle of its own resources w/o using any of the kube-burner APIs.

Summarizing, I'm not against of merging this PR, since we need this functionality ASAP, and its not affecting any other workload, but I sincerely think that in the future we should consider taking the bits of this workload out of `kube-burner-ocp`.

vishnuchalla:
> As we already discussed in slack, this workload doesn't make an extensive usage of the kube-burner functionalities , it's more similar to a script than an actual workload. Most of actions are executed in a not `kube-burner` way, as the scripted actions of the logic already take care of the lifecycle of its own resources w/o using any of the kube-burner APIs.
> 
> Summarizing, I'm not against of merging this PR, since we need this functionality ASAP, and its not affecting any other workload, but I sincerely think that in the future we should consider taking the bits of this workload out of `kube-burner-ocp`.

@rsevilla87 Thanks for approving these changes. 
Well, its a mix of things when it comes to functionalities. Core logic is handled with `openshift/client-go` and still uses kube-burner functionalities like scraping prometheus, es-indexing also some common utils.
Regarding `scripts` vs `workflow`, as long is it is decoupled and easy to re-use and maintain, I am okay. IMHO, managing/maintaining this with shell scripts is lot more complex in terms of operational overhead without any added benefit. Also there is a high chance for it break our existing workflows often in CI. 
Alongside, I see this PR as an opportunity to integrate `openshift/client-go` which will be more like a `ocp-wrapper` thing helping us to interact with some of the `openshift` specific resources. We should also be open to add other openshift related to this repo which isn't tightly coupled with kube-burner's execution flow in specific.

I have more refactoring/restructuring coming up in my follow up PRs. Hoping that to make things more clear.

PR #103, opened 2024-09-11
add more stackrox/acs metrics

davdhacs:
Copied from https://github.com/stackrox/stackrox/blob/master/tests/performance/scale/tests/kube-burner/cluster-density/metrics.yml


cmd/config/metrics-report.yml
```diff
@@ -224,6 +224,788 @@
   metricName: max-memory-stackrox
   instant: true
 
+# ACS - Central
```
rsevilla87, line 227: I don't think we should add ACS metrics in this file, metrics here are for standard OCP deployments, how are you using it? There's not any ACS related workload in this repository

rsevilla87: If you're using this metrics file from an external repo I recommend you to create a new metrics file with all the acs/stackrox stuff rather than modifying this one

PR #111, opened 2024-09-25
Custom metric-profile list from flag

### LINKED ISSUE #110: [RFE] Set custom metric-profiles from flag

With the current implementation, each workload has predefined a metric-profile (metrics.yml, metrics-aggregated.yml, metrics-egress.yml, etc), it would be useful to allow the user to use an external metric-profile

A metric-profile list can be passed as a flagThat metric profile can be passed from a flag, i.e:

```shell
 ./bin/amd64/kube-burner-ocp cluster-density-v2 --iterations=10 --metrics-profile=custom-metrics.yml,more-metrics.yml
```

If this flag is not set, kube-burner will follow the previous behavior, of using the pre-defined embedded metrics-profile of the workload in question.

**Describe alternatives you've considered**

As part of this effort, we can end up deprecating the flag `--profile-type`, that would mean that the files specified by `--metrics-profile` can also be embedded in the kube-burner-ocp binary

**Additional context**

Having this PR can help to fix issues like the one that is trying to fix with https://github.com/kube-burner/kube-burner-ocp/pull/103, which basically is extending  the default metrics profile with not relevant prometheus expresions for standards clusters due to the impossibility of using a different metrics-profile rather than the default one.

rsevilla87:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description

Small implementation of the --metrics-profile flag

- Related Issue #
- Closes #110

### DIFF
--- a/common.go
+++ b/common.go
@@ -28,16 +28,13 @@ import (
 
 var clusterMetadata ocpmetadata.ClusterMetadata
 
-func setMetrics(cmd *cobra.Command, metricsProfile string) {
-	var metricsProfiles []string
+func setMetrics(cmd *cobra.Command, metricsProfiles []string) {
 	profileType, _ := cmd.Root().PersistentFlags().GetString("profile-type")
 	switch ProfileType(profileType) {
 	case Reporting:
 		metricsProfiles = []string{"metrics-report.yml"}
-	case Regular:
-		metricsProfiles = []string{metricsProfile}
 	case Both:
-		metricsProfiles = []string{"metrics-report.yml", metricsProfile}
+		metricsProfiles = append(metricsProfiles, "metrics-report.yml")
 	}
 	os.Setenv("METRICS", strings.Join(metricsProfiles, ","))
 }
--- a/cluster-density.go
+++ b/cluster-density.go
@@ -33,6 +33,7 @@ func NewClusterDensity(wh *workloads.WorkloadHelper, variant string) *cobra.Comm
 	var churnDelay, churnDuration time.Duration
 	var churnDeletionStrategy string
 	var podReadyThreshold time.Duration
+	var metricsProfiles []string
 	cmd := &cobra.Command{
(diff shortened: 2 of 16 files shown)


rsevilla87:
> This looks like it it exactly what I need. Thank you!

Great, have you tested it?

jtaleric (commented):
lgtm - @davdhacs can you confirm you have tested and this fits the needs for the ACS team?

common.go
```diff
@@ -28,16 +28,13 @@ import (
...
 var clusterMetadata ocpmetadata.ClusterMetadata
 
-func setMetrics(cmd *cobra.Command, metricsProfile string) {
-	var metricsProfiles []string
+func setMetrics(cmd *cobra.Command, metricsProfiles []string) {
 	profileType, _ := cmd.Root().PersistentFlags().GetString("profile-type")
 	switch ProfileType(profileType) {
 	case Reporting:
 		metricsProfiles = []string{"metrics-report.yml"}
-	case Regular:
```
vishnuchalla, line 37: We might want to remove this from the help message in `ocp.go` as well

rsevilla87: I don't want to touch any of the current funtionallity, for now, in a follow-up PR my idea is to deprecate the profile-type flag too

vishnuchalla (approved):
lgtm

common.go
vishnuchalla: Without `regular` case being handled here, there is no point in having it in the [help description](https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/ocp.go#L62) as its is dead code and might be misleading for users.
Can you provide more details on deprecating profile-type flag? It might break our CI.

rsevilla87: it will break it, for sure, for that reason I prefer to address that separately

PR #112, opened 2024-09-25
Remove useless flags

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

These flags and vars have no any effect and can be removed to avoid confussions

- Related Issue #
- Closes #

### DIFF
--- a/udn-density-l3-pods.go
+++ b/udn-density-l3-pods.go
@@ -17,19 +17,17 @@ package ocp
 import (
 	"fmt"
 	"os"
-	"strconv"
 	"time"
 
 	"github.com/kube-burner/kube-burner/pkg/workloads"
-	log "github.com/sirupsen/logrus"
 
 	"github.com/spf13/cobra"
 )
 
 // NewUDNDensityL3Pods holds udn-density-l3-pods workload
 func NewUDNDensityL3Pods(wh *workloads.WorkloadHelper) *cobra.Command {
-	var churnPercent, churnCycles, iterations, podsPerNode int
-	var churn, svcLatency bool
+	var churnPercent, churnCycles, iterations int
+	var churn bool
 	var churnDelay, churnDuration, podReadyThreshold time.Duration
 	var churnDeletionStrategy string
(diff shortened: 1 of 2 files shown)


udn-density-l3-pods.go
```diff
@@ -65,7 +57,5 @@ func NewUDNDensityL3Pods(wh *workloads.WorkloadHelper) *cobra.Command {
 	cmd.Flags().StringVar(&churnDeletionStrategy, "churn-deletion-strategy", "default", "Churn deletion strategy to use")
 	cmd.Flags().IntVar(&iterations, "iterations", 0, "Iterations")
 	cmd.Flags().DurationVar(&podReadyThreshold, "pod-ready-threshold", 1*time.Minute, "Pod ready timeout threshold")
-	cmd.Flags().IntVar(&podsPerNode, "pods-per-node", 245, "Pods per node")
-	cmd.Flags().BoolVar(&svcLatency, "service-latency", false, "Enable service latency measurement")
```
krishvoor, line 69: I do not think udn-l3 workload supports churning (yet) - can we disable it?
`cmd.Flags().BoolVar(&churn, "churn", false, "disable churning")`

josecastillolema: It looks like it does: https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/udn-density-l3-pods/udn-density-l3-pods.yml#L44-L49

krishvoor: @venkataanil Can you confirm if you will be using Churning?

krishvoor: > It looks like it does: https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/udn-density-l3-pods/udn-density-l3-pods.yml#L44-L49

IIUC - we might not have immediate/no value with churning, but if it's otherwise I'm ok to leave it as-is.

PR #116, opened 2024-10-01
Removing Dev ES instance reference

vishnuchalla:
## Type of change

- [x] Refactor
- [ ] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Removing dev ES instance and replacing it with prod ones

- Closes # https://issues.redhat.com/browse/PERFSCALE-3430

## Testing
Can only be tested once this PR is merged.

### DIFF
--- a/.github/workflows/ci-tests.yml
+++ b/.github/workflows/ci-tests.yml
@@ -1,6 +1,7 @@
 name: CI tests
 on:
   workflow_call:
+  workflow_dispatch:
   pull_request_target:
     types: [opened, edited, synchronize, reopened, labeled]
     branches:
--- a/.github/workflows/test-ocp.yml
+++ b/.github/workflows/test-ocp.yml
@@ -1,5 +1,6 @@
 name: Execute tests on OCP
 on:
+  workflow_dispatch:
   workflow_call:
     secrets:
       OPENSHIFT_SERVER:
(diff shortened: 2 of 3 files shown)


.github/workflows/ci-tests.yml
```diff
@@ -12,11 +12,9 @@ jobs:
     uses: ./.github/workflows/linters.yml
 
   build:
-    if: ${{ contains(github.event.pull_request.labels.*.name, 'ok-to-test') }}
```
rsevilla87, line 15: why are you removing this?

vishnuchalla: After this change, CI tests will only be executed after a merge in main. So this check won't be required.

rsevilla87: I don't understand, the trigger event is still pull_request_target

rsevilla87: Also, why do you want that change?, secrets can be acceded from a PR too

vishnuchalla: Yes you are right. According to the docs [here](https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target), `pull_request_target` seems to be providing gh-secret access to the fork PRs (for example, openshift cluster secrets working against PRs). Not having a way to [test it](https://github.com/kube-burner/kube-burner-ocp/actions/runs/11223524706/job/31198394006?pr=116) before the merge confused me. 

So technically after this PR is merged, the follow up PRs should be able to succeed on the current failing CI test.
While `pull_request_target` provides untrusted access to the forks, `pull_request` will help providing access to the PRs only against the main stream repo not for forks. Previous experiment [here](https://github.com/cloud-bulldozer/orion/actions/runs/10304354144)

vishnuchalla: Changed and updated the PR as suggested.

rsevilla87 (approved):
lgtm

PR #118, opened 2024-10-09
Arrange deployments that depend on secrets and configmaps to be created after said resources

afcollins:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

The cluster-density deployments are created before the secrets and configmaps they depend on.
This can cause scheduling delays and in some cases cause errors in the apiserver.

- Related Issue #
- Closes #

- [x] If it is a core feature, I have added thorough tests.

### DIFF
--- a/cmd/config/cluster-density-ms/cluster-density-ms.yml
+++ b/cmd/config/cluster-density-ms/cluster-density-ms.yml
@@ -53,11 +53,6 @@ jobs:
       - objectTemplate: imagestream.yml
         replicas: 1
 
-      - objectTemplate: deployment.yml
-        replicas: 4
-        inputVars:
-          podReplicas: 2
-
       - objectTemplate: service.yml
         replicas: 2
         
@@ -69,3 +64,8 @@ jobs:
 
       - objectTemplate: configmap.yml
         replicas: 10
+
+      - objectTemplate: deployment.yml
+        replicas: 4
+        inputVars:
+          podReplicas: 2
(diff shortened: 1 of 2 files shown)


afcollins:
As @shashank-boyapally suspected, it does reduce our Ready latencies:

```
time="2024-10-09 15:54:26" level=info msg="cluster-density-v2: Ready 99th: 11000 max: 11000 avg: 3816" file="pod_latency.go:237"
```

shashank-boyapally:
Hi @afcollins, Looks like in that case we might need to redefine our thresholds.. cc: @rsevilla87 @vishnuchalla

rsevilla87:
> As @shashank-boyapally suspected, it does reduce our Ready latencies:
> 
> ```
> time="2024-10-09 15:54:26" level=info msg="cluster-density-v2: Ready 99th: 11000 max: 11000 avg: 3816" file="pod_latency.go:237"
> ```

what was the 99th original value?

rsevilla87:
> > what was the 99th original value?
> 
> 13 seconds is the minimum with the current ordering.

interesting, I wonder how much would be the difference in a larger scenario like the ones in CPT

afcollins:
Yeah, curious what it will be.

Hopefully we can agree since the latency has decreased that it won't break anything and we can merge it.

vishnuchalla (approved):
lgtm

PR #139, opened 2024-10-31
Disable heatlh-check by default

jtaleric:
## Type of change

- [ ] Refactor
- [ ] New feature
- [x] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Some aspects of kube-burner do not need to have health-check, such as the kube-burner-ocp index command.

- Related Issue #
- Closes #

### DIFF
--- a/cmd/ocp.go
+++ b/cmd/ocp.go
@@ -50,7 +50,7 @@ func openShiftCmd() *cobra.Command {
 	ocpCmd.PersistentFlags().BoolVar(&localIndexing, "local-indexing", false, "Enable local indexing")
 	ocpCmd.PersistentFlags().StringVar(&workloadConfig.MetricsEndpoint, "metrics-endpoint", "", "YAML file with a list of metric endpoints, overrides the es-server and es-index flags")
 	ocpCmd.PersistentFlags().BoolVar(&alerting, "alerting", true, "Enable alerting")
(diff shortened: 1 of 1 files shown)


rsevilla87:
I think this PR is overkill, health check is being used currently w/o specifying it in out CPT pipelines, if you want to skip it in the `kube-burner-ocp index` command you can just update this conditional to also include index

https://github.com/kube-burner/kube-burner-ocp/blob/d2588e055430f7de5f0f26d2407cf9cc658e9bbd/cmd/ocp.go#L76-L78

jtaleric:
> I find this PR a bit overkill, health check is being used currently in out CPT pipelines (w/o enabling it specifically), if you want to skip it in the `kube-burner-ocp index` command you can just update this conditional to also include index
> 
> https://github.com/kube-burner/kube-burner-ocp/blob/d2588e055430f7de5f0f26d2407cf9cc658e9bbd/cmd/ocp.go#L76-L78

I would argue if we should even have a health-check given there are other tools which do the exact same thing, like `oc adm wait-for-cluster-stable` 

I prefer the oc method as if you run `health-check` which is fine for that point in time -- the oc command actually has a option to make sure the cluster is stable for > N number of minutes. 

However, understood. I think we could also just update the workflow to `--check-health=False`

PR #140, opened 2024-11-04
Adding l2 udn workload

jtaleric:
## Type of change

- [ ] Refactor
- [x] New feature
- [ ] Bug fix
- [ ] Optimization
- [ ] Documentation Update

## Description
Eventually we should consider having a `udn-density-pods` workload and allow the user to choose L3 or L2.

- Related Issue #
- Closes #


udn-density-pods.go
```diff
@@ -47,12 +47,23 @@ func NewUDNDensityL3Pods(wh *workloads.WorkloadHelper) *cobra.Command {
 		},
 		Run: func(cmd *cobra.Command, args []string) {
 			setMetrics(cmd, metricsProfiles)
-			rc = wh.Run(cmd.Name())
+			if l2 && l3 {
+				// unset l3 if both are enabled.
+				l3 = false
```
rsevilla87, line 52: You can use the cobra function cmd.MarkFlagsMutuallyExclusive() to avoid this check

rsevilla87: https://github.com/kube-burner/kube-burner-ocp/blob/762ae83e3c9fee8d8c91515bbfa96ccc07df41af/custom-workload.go#L90-L91

cmd/config/udn-density-l2-pods/udn.yml
```diff
@@ -0,0 +1,10 @@
+
+apiVersion: k8s.ovn.org/v1
+kind: UserDefinedNetwork
+metadata:
+  name: l2-network-{{.Iteration}}
+spec:
+  topology: Layer2
+  layer2:
+      role: Primary
+      subnets: ["10.132.0.0/16"]
```
rsevilla87, line 10: Do we want to parameterize this CIDR?

udn-density-pods.go
jtaleric: I still need this condition -- because I want to default to l3 (to avoid having to change what we have in CI right now)... so if the user passes layer2, I will just unset layer3.

cmd/config/udn-density-l2-pods/udn.yml
rsevilla87: ok, just double checking

rsevilla87 (approved):
lgtm

PR #149, opened 2024-11-19
add connectivity check via svc in readiness probe

mohit-sheth:
## Description

Adding readiness probe which tries to connect to the server pod via the service endpoint
Routes support is not in yet for UDN


rsevilla87 (commented):
You want to add something similar to l2 too?

mohit-sheth:
> You want to add something similar to l2 too?

now the l2 and l3 parts are merged so it should be good

jtaleric (approved):
/lgtm

vishnuchalla (approved):
lgtm

vishnuchalla:
@mohit-sheth Do you have plans to add a functional test to this workload in the `test` directory. We do not seem to be having one as of today.

vishnuchalla:
> @mohit-sheth Do you have plans to add a functional test to this workload in the `test` directory. We do not seem to be having one as of today.

Merging as per the conversation offline, to add tests after 4.18 release as this requires a featuregate.

PR #161, opened 2024-12-10
Disable web-burner CI tests

rsevilla87:
## Type of change

- [x] Refactor

## Description

These CI tests don't work in the new CI cluster deployed in BM


rsevilla87:
web-burner CI works in the new cluster, there's no need to disable it

PR #141, opened 2024-11-05
Decrease readinessProbe period

rsevilla87:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation Update

## Description

Decreasing this period should improve the measurement accuracy.

## Related Tickets & Documents

- Related Issue #
- Closes #

## Checklist before requesting a review

- [x] I have performed a self-review of my code.
- [ ] If it is a core feature, I have added thorough tests.

PR #163, opened 2024-12-11
Revert "Decrease readinessProbe period"

rsevilla87:
Reverts kube-burner/kube-burner-ocp#141


afcollins (approved):
LGTM

I think 2s may be a little low for the server pods to be ready and we are getting way more probe failures than waiting for 10s.

There could be middle ground around 5s, but also use of `initialDelaySeconds=5` may allow us to get finer-resolution than 5s period while also not over-burdening when the server is not yet available

PR #171, opened 2025-01-13
Split udn creation and other objects, parameterize job pause

mohit-sheth:
## Type of change

- Optimization

## Description
We want to run two cases

- udn and pods are created simultaneously in a namespaces, this is good for stressing the CNI
- wait for udns to come up and then create pods, a more practical approach

to switch between two cases we can simply change the ` --job-pause` time to 1s for case1 or X mins, X depends on how much time it takes to create udns.

Tested locally
```
./bin/amd64/kube-burner-ocp udn-density-pods --iterations=2 --gc=false --churn=false --job-pause 1s
time="2025-01-13 17:10:41" level=info msg="❤️ Checking for Cluster Health" file="cluster_health.go:46"
time="2025-01-13 17:10:41" level=info msg="Cluster is Healthy" file="cluster_health.go:54"
time="2025-01-13 17:10:41" level=info msg="Layer 3 is enabled" file="udn-density-pods.go:54"
<snip>
time="2025-01-13 17:11:47" level=info msg="Verifying created objects" file="utils.go:92"
time="2025-01-13 17:11:47" level=info msg="Pausing for 1s before finishing job" file="job.go:153"
```

### DIFF
--- a/cmd/config/udn-density-pods/udn-density-pods.yml
+++ b/cmd/config/udn-density-pods/udn-density-pods.yml
@@ -28,10 +28,41 @@ metricsEndpoints:
 
 jobs:
   {{ if eq .ENABLE_LAYER_3 "true" }}
-  - name: udn-density-l3-pods
+  - name: create-udn-l3
   {{ else }}
-  - name: udn-density-l2-pods
+  - name: create-udn-l2
   {{ end }}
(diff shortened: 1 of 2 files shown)


jtaleric (approved):
/lgtm

PR #154, opened 2024-11-25
[PERFSCALE-3598] virt-density

rsevilla87:
## Type of change

- [x] New feature

## Description

New virt-density workload
cc @jeniferh 

## Related Tickets & Documents

- Related Issue #
- Closes #

PR #172, opened 2025-01-16
[PERFSCALE-3598] Virt-density test

josecastillolema:
## Type of change

- CI

## Description

Test for the virt-density workload (https://github.com/kube-burner/kube-burner-ocp/pull/154)


josecastillolema:
@jtaleric @vishnuchalla should we export the results to the ES instance? Thoughts?

josecastillolema:
@vishnuchalla any idea why **tests/ocp-e2e-ci** is forever stuck on _Expected - Waiting for status to be reported_ ?
Does it need any special labels like `ok-to-test` ?
Observed the same behavior on https://github.com/kube-burner/kube-burner-ocp/pull/154

krishvoor:
@josecastillolema: Yup that's needed does, someone from the org + non-author needs to apply the label
>Does it need any special labels like ok-to-test ?

jtaleric:
> @jtaleric @vishnuchalla should we export the results to the ES instance? Thoughts?

I don't think that is necessary as we already have issues with storing ES results (due too many metrics). Maybe we can use a local indexer and check if specific metrics the workload captures with regex?

test/test-ocp.bats
```diff
@@ -92,6 +92,10 @@ teardown_file() {
   run_cmd kube-burner-ocp crd-scale --iterations=10 --alerting=false
 }
 
+@test "virt-density" {
```
vishnuchalla, line 95: @josecastillolema Can you get some checks added similar to [this](https://github.com/kube-burner/kube-burner-ocp/blob/e99168445336c4d75b2fea28e3f8d63fe101e845/test/test-ocp.bats#L55)?

josecastillolema: What metrics are we looking for? Looking at https://github.com/kube-burner/kube-burner-ocp/pull/154 I dont see any new metric here specific to this use case

vishnuchalla: `vmiLatencyMeasurement`, `vmiLatencyQuantilesMeasurement`, `jobSummary`, `etcdVersion` etc. Runtime measurements and some key metrics in the metrics profile being used are good to consider while performing checks.

josecastillolema: OK! So I added some of them (_vmiLatencyMeasurement_ and _vmiLatencyQuantilesMeasurement_).
Now the test is failing:
```
    time="2025-01-26 18:47:10" level=info msg="Stopping measurement: vmiLatency" file="factory.go:129"
 ...
   time="2025-01-26 18:47:25" level=info msg="👋 kube-burner run completed with rc 0 for UUID 901f2323-9f02-4b5e-8764-5a9c2912943f" file="helpers.go:82"
   vmiLatencyMeasurement not found in ***/kube-burner-ocp/_search?q=uuid.keyword:238cca7f-34ec-4e7f-b4dd-5fd2a8c0cdbb+AND+metricName.keyword:vmiLatencyMeasurement+AND+(metadata.ocpVersion.keyword:*+OR+ocpVersion.keyword:*)
```

Why is looking for UUID 238cca7f-34ec-4e7f-b4dd-5fd2a8c0cdbb instead of 901f2323-9f02-4b5e-8764-5a9c2912943f? Am I missing smth?

vishnuchalla: It must have been looking for an UUID that is exposed somewhere else in the test suite. To fix this. can you try passing UUID as a parameter to kube-burner-ocp command ?

josecastillolema: Thanks @vishnuchalla !
This fixed the UUID mismatch indeed.
However, still no luck:
```
...
time="2025-01-28 09:56:14" level=info msg="👋 kube-burner run completed with rc 0 for UUID 39b3fd10-979e-4540-9ce8-96c3e15fe991" file="helpers.go:82"
   vmiLatencyMeasurement not found in ***/kube-burner-ocp/_search?q=uuid.keyword:39b3fd10-979e-4540-9ce8-96c3e15fe991+AND+metricName.keyword:vmiLatencyMeasurement+AND+(metadata.ocpVersion.keyword:*+OR+ocpVersion.keyword:*)
```

cmd/ocp.go
```diff
@@ -94,7 +94,7 @@ func openShiftCmd() *cobra.Command {
 			envVars["ALERTS"] = ""
 		}
 		// If metricsEndpoint is not set, use values from flags
-		if workloadConfig.MetricsEndpoint == "" {
+		if workloadConfig.MetricsEndpoint == "" && esServer != "" && esIndex != "" {
```
josecastillolema, line 97: Bug?

vishnuchalla: Its not exactly a bug, but an additional check to prevent overrides.

vishnuchalla:
@josecastillolema now that you have come to know the issue, can you squash your commits using my code changes and make DCO happy?

PR #174, opened 2025-01-23
add netpols to udn workload

mohit-sheth:
## Type of change

- Optimization

## Description
Adding 2 networkpolicies which first deny-all traffic and then allow client->server communication within the namespaces/UDN


vishnuchalla (approved):
lgtm

PR #177, opened 2025-01-24
Makefile - Find the local arch if not provided

ygalblum:
## Type of change

- Optimization

## Description

If ARCH is not set, use the machine arch and not force amd64


vishnuchalla (approved):
lgtm

PR #180, opened 2025-01-28
Add virt capacity benchmark test

ygalblum:
## Type of change

- New feature

## Description

Run Virtualization with Storage Capacity Benchmark test

### DIFF
--- a/cmd/config/virt-capacity-benchmark/virt-capacity-benchmark.yml
+++ b/cmd/config/virt-capacity-benchmark/virt-capacity-benchmark.yml
@@ -1,0 +1,214 @@
+{{- $kubeBurnerFQDN := "kube-burner.io" -}}
+{{- $testName := "virt-capacity-benchmark" }}
+{{- $nsName := .testNamespace -}}
+{{- $vmCount := .vmCount -}}
+{{- $vmName := $testName -}}
+{{- $sshPublicKeySecretName := $testName -}}
+{{- $rootVolumeSizeStr := (list (.rootVolumeSize | toString) "Gi") | join "" -}}
+{{- $dataVolumeSizeStr := (list (.dataVolumeSize | toString) "Gi") | join "" -}}
+{{- $resizedRootVolumeSize := add .rootVolumeSize .volumeSizeIncrement -}}
+{{- $resizedDataVolumeSize := add .dataVolumeSize .volumeSizeIncrement -}}
+{{- $volumeLabelKey := (list $testName "." $kubeBurnerFQDN "/volume-type") | join "" -}}
+{{- $volumeLabelValueRoot := "root" -}}
+{{- $volumeLabelValueData := "data" -}}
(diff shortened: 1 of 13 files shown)


common.go
```diff
@@ -70,3 +90,175 @@ func GatherMetadata(wh *workloads.WorkloadHelper, alerting bool) error {
 	}
 	return nil
 }
+
+// GenerateSSHKeyPair generates an SSH key pair and saves them to the specified files
+func generateSSHKeyPair(sshKeyPairPath string) (string, string, error) {
```
vishnuchalla, line 95: Can this go into `go-commons` please? Looks like nice function to be reused across our tooling.

common.go
```diff
@@ -70,3 +90,175 @@ func GatherMetadata(wh *workloads.WorkloadHelper, alerting bool) error {
...
+
+func getVolumeSnapshotClassNameForStorageClass(storageClassName string) (string, error) {
+	provisioner, err := getStorageClassProvisioner(storageClassName)
+	if err != nil {
+		return "", err
+	}
+	return getVolumeSnapshotClassNameForProvisioner(provisioner)
+}
+
+func virtctlExists() bool {
```
vishnuchalla, line 261: Some of the functions in this file seem to be commonly used across all the virtualization portifolio. Can we move all of them to go-commons please?

ygalblum:
I've updated the code. However, it now depends on changes made in https://github.com/cloud-bulldozer/go-commons/. Once the all changes are merged and the dependency is updated here, this code change will compile successfully

cmd/config/virt-capacity-benchmark/virt-capacity-benchmark.yml
```diff
@@ -0,0 +1,207 @@
...
+  objects:
+  - kind: Namespace
+    labelSelector:
+      {{ $testNamespacesLabelKey }}: {{ $testNamespacesLabelValue }}
+{{ end }}
+
+- name: create-vms
+  jobType: create
+  jobIterations: 1
+  qps: 20
```
jtaleric, line 44: curious why in the first job you have the CNV default of 5/10 for QPS/Burst -- but here and further down you default to 20/20?

ygalblum: TBH I didn't put too much thought into it. Especially since most of the jobs are pretty small. So, not sure it even matters. Which values would you recommend here?

ygalblum:
@vishnuchalla I've added nobs that allow skipping some checks and setting some values so that the test can pass with the CI cluster. Hopefully the test will pass in the CI as well (along with the checks I've added)

ygalblum:
The CI is failing because the `BeforeCleanup` script is implemented for `bash` while the default shell in Ubuntu (which is the runner) is `dash`. I'm looking into how to address this. Probably need to address it in `kube-burner`

ygalblum:
@vishnuchalla PTAL. I solved the CI issue by changing the script so that it will work also on `dash`.

vishnuchalla (approved):
lgtm

PR #182, opened 2025-01-29
Add the required label to primary UDN namespaces

kyrtapz:
Primary UDNs require the namespaces to be labeled with a known label.


jtaleric (approved):
/lgtm

PR #184, opened 2025-02-05
add ovn pprof collection to control plane workloads

mohit-sheth:
## Type of change
- Optimization

## Description
pprofs are almost all the times required to debug a regression in CPU utilization
When our prow runs show this regression we should supplement our logs with these pprofs

### DIFF
--- a/cmd/config/cluster-density-v2/cluster-density-v2.yml
+++ b/cmd/config/cluster-density-v2/cluster-density-v2.yml
@@ -8,6 +8,24 @@ global:
         - conditionType: Ready
           metric: P99
           threshold: {{.POD_READY_THRESHOLD}}
+{{ if eq .PPROF "true" }}
+    - name: pprof
+      pprofInterval: 2m
+      pprofDirectory: pprof-data
+      pprofTargets:
+      - name: ovnkube-controller
+        namespace: "openshift-ovn-kubernetes"
+        labelSelector: {app: ovnkube-node}
+        url: http://localhost:29103/debug/pprof/profile?seconds=30
+      - name: ovn-controller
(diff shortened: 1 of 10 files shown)


cmd/config/cluster-density-v2/cluster-density-v2.yml
```diff
@@ -8,6 +8,14 @@ global:
...
           metric: P99
           threshold: {{.POD_READY_THRESHOLD}}
+    - name: pprof
+      pprofInterval: 2m
+      pprofDirectory: pprof-data
+      pprofTargets:
+      - name: ovn-profile
+        namespace: "openshift-ovn-kubernetes"
+        labelSelector: {app: ovnkube-node}
+        url: http://localhost:29103/debug/pprof/profile?seconds=30
```
krishvoor, line 18: Can we toggle via `pprof_needed=yes`, so we don't always default to collect

mohit-sheth: good point, updated

jtaleric: should we have all the ovn components vs just ovnk-node?

mohit-sheth: added ocn-controller and ovnk-control plane

vishnuchalla:
we need to resolve conflicts. Apart from that, I wonder if pprof collection itself are going to impact our current cpu/memory usage metrics. If there is a way to distinguish them from our metrics then it should be good to go.

krishvoor (approved):
/lgtm

venkataanil (approved):
LGTM

jtaleric (approved):

PR #192, opened 2025-02-12
Standardizing on fedora VM image

jeniferh:
## Type of change

- Optimization

## Description
Cirros has limited usage so standardizing on Fedora, and we agreed 1Gi is a good min VM size


jeniferh:
cc @ygalblum any concerns w/ this change?

ygalblum:
> cc @ygalblum any concerns w/ this change?

I don't mind the move to fedora. I would maybe consider adding a parameter to the test to allow the user to change the image. But, not a must.

As for the RAM size. I guess the question is what are we testing here. If just the creation of VMs, will this pose a limit on the amount of VMs?

jeniferh:
> I don't mind the move to fedora. I would maybe consider adding a parameter to the test to allow the user to change the image. But, not a must.

Hmm yeah good point, an image parameter would be nice. 
> 
> As for the RAM size. I guess the question is what are we testing here. If just the creation of VMs, will this pose a limit on the amount of VMs?

Technically we only support min 512MB (https://access.redhat.com/articles/6571671) although we know Cirros can boot with less. It is a scheduling constraint, hmm again maybe a param is nice :)

ygalblum:
> Technically we only support min 512MB (https://access.redhat.com/articles/6571671) although we know Cirros can boot with less. It is a scheduling constraint, hmm again maybe a param is nice :)

Yeah, I didn't want to say parameter twice :). But, even with a parameter, there should be a default value. Either 512Mi or 1Gi. Whatever you think is best.
BTW I added a simple way to pass values from `kube-burner-ocp` to `kube-burner` using a map instead of through environment variables: https://github.com/kube-burner/kube-burner/blob/79192395a592d516ae73d25bd94af514a060b14f/pkg/workloads/helpers.go#L51

capolrik (approved):

jeniferh:
Thanks @vishnuchalla , I've been endlessly distracted... but I did open a backlog story to address Ygal's suggestions: https://issues.redhat.com/browse/PERFSCALE-3809

vishnuchalla:
> Thanks @vishnuchalla , I've been endlessly distracted... but I did open a backlog story to address Ygal's suggestions: https://issues.redhat.com/browse/PERFSCALE-3809

Thank you.

PR #194, opened 2025-02-21
Add support for multinetworkpoliy

venkataanil:
User creates a definite number of vlans (nncp) and then uses them across namespaces using NADs.
multinetworkpoliy used for additional interfaces of pods and virtual machines.
Netpol latency measurment will be implemented later based on if multinetworkpoliy created for pods or virtual machines.

This workload uses the same config options like regular network policy, however
1) selector based approach not possible for spec.podselector and
 ingress.from.podselctor for the virtual machines
2) in regular network policy, any namespace can be chosen as peer,
 however now the namespace within the same vlan has to be chosen
 as peer.


vishnuchalla:
@venkataanil can we get a small CI test added please?

mohit-sheth:
thanks @venkataanil for a quick turnaround

cmd/config/network-policy/nad-vlan.yml
```diff
@@ -0,0 +1,18 @@
+{{- $vlanId := add (div $.Iteration .namespaces_per_vlan) 1}}
```
capolrik, line 1: Is there a world where vlanId can be higher than 255 and then generates error in the JSON config?

cmd/config/network-policy/vm-localnet.yml
```diff
@@ -0,0 +1,49 @@
...
+          - name: br-ex-vlan-{{$vlanId}}
+            bridge: {}
+      networks:
+      - multus:
+          networkName: br-ex-vlan-{{$vlanId}}
+        name: br-ex-vlan-{{$vlanId}}
+      volumes:
+      - name: containerdisk
+        containerDisk:
+          image: quay.io/openshift-cnv/qe-cnv-tests-fedora:40
```
capolrik, line 39: Maybe it is worth using a variable here to easily change the VM image.

capolrik (commented):
Really nice! I have been able to reproduce the pods and VMs scenario on my cluster and here are some feedback:

- It may be worth mentioning the requirements before launching this scenario like nmstate operator should be installed, multi-networkpolicy CRD should be present...
- When looking at the jobs in grafana Kube-burner report, you don't see the difference in the metadata between multi-network enabled/disabled and the same for pods/vms you don't know which one has been used.

rsevilla87:
Is this PR still relevant @venkataanil

PR #195, opened 2025-02-24
CI - Install virtctl

ygalblum:
## Type of change

- CI

## Description

Install virtctl since it is a requirement for the tests


ygalblum:
virtctl is a requirement in for the tests added in https://github.com/kube-burner/kube-burner-ocp/pull/180. Adding it here before so that the test can run after a rebase

PR #197, opened 2025-02-26
Add test for creating Virtual Machines from a master image

ygalblum:
## Type of change

- New feature

## Description

The new test follows the VDI scenario in which VMs are created based on a master image from another VM

### DIFF
--- a/cmd/config/virt-clone/check.sh
+++ b/cmd/config/virt-clone/check.sh
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+COMMAND=$1
+LABEL_KEY=$2
+LABEL_VALUE=$3
+NAMESPACE=$4
+IDENTITY_FILE=$5
+REMOTE_USER=$6
+EXPECTED_ROOT_SIZE=$7
+EXPECTED_DATA_SIZE=$8
+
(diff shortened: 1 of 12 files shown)


vishnuchalla (approved):
lgtm

cmd/config/virt-clone/check.sh
```diff
@@ -0,0 +1,107 @@
...
+    vms=$(kubectl get vm -n "${namespace}" -l "${label_key}"="${label_value}" -o json | jq .items | jq -r '.[] | .metadata.name')
+    local ret=$?
+    if [ $ret -ne 0 ]; then
+        echo "Failed to get VM list"
+        exit 1
+    fi
+    echo "${vms}"
+}
+
+remote_command() {
```
vishnuchalla, line 39: Getting this merged to unblock you, but I think having some of these utilities in go-commons would be nice for re-usability.

ygalblum: I'm not sure this is doable for adding shell scripts into the embedded filesystem. The way this currently works requires me to duplicate the script for each workload.
See https://github.com/kube-burner/kube-burner/pull/783 for a similar issue

PR #198, opened 2025-02-28
deprecate network-policy-v1 test cases

mohit-sheth:
## Type of change

- Refactor

## Description
Deprecate networkpolicy v1 wokload in favour of new network-policy workload https://github.com/kube-burner/kube-burner-ocp/tree/main#network-policy-workloads


smalleni (approved):
LGTM

vishnuchalla:
@mohit-sheth config is removed which is good. But there a bit of GO source code too which will be dead after this removal. Mind cleaning up that as well please?

Issue #203, opened 2025-03-10
[BUG] ES_SERVER variable rendering issue in workload templates


vishnuchalla:
# Bug Description
When ES_SERVER is unset we see this failure even if we do not want any ES indexing
```
time="2025-03-07 05:21:19" level=fatal msg="error rendering configuration template: rendering error: template: :34:6: executing \"\" at <.ES_SERVER>: map has no entry for key \"ES_SERVER\"" file="helpers.go:65"
```

## **Output of `kube-burner-ocp` version**
```
Version: main
Git Commit: 8eeb7149e021354be48ebdc510adbcd47f7b497c
Build Date: 2025-03-06-19:56:03
Go Version: go1.23.4
OS/Arch: linux amd64
```

## **Describe the bug**
Template rendering runs into errors when ES_SERVER environment variable is unset.

## **To Reproduce**

1. Build and install kube-burner-ocp wrapper binary.
2. Make sure `ES_SERVER` env is unset. `unset ES_SERVER`.
3. Then try running any workload. For example 
```
vchalla@vchalla-thinkpadp1gen2:~/myforks/kube-burner-ocp$ kube-burner-ocp cluster-density-v2 --iterations=180 --churn=False --alerting=False --timeout=40m 
time="2025-03-10 13:28:44" level=info msg="❤️ Checking for Cluster Health" file="cluster_health.go:46"
time="2025-03-10 13:28:45" level=info msg="Cluster is Healthy" file="cluster_health.go:54"
time="2025-03-10 13:28:46" level=fatal msg="error rendering configuration template: rendering error: template: :34:6: executing \"\" at <.ES_SERVER>: map has no entry for key \"ES_SERVER\"" file="helpers.go:65"
```

## **Expected behavior**
The error shouldn't appear and shouldn't block the workload from running. It should ideally proceed without any indexing as local indexing isn't enabled either.

PR #208, opened 2025-03-15
Lower barrier of entry by making indexing truly optional

smalleni:
Currently the go template checks that ES_SERVER is non empty and would throw a rendering error when the variable is not defined. While the intent has always been to make indexing optional, the current implementation does not support that.

We would like to enable cases where users do not care about ES or local indexing and merely want to run load quickly by using something like `kube-burner-ocp node-density` use the stdout to read high level measurements like pod ready latency.

## Type of change

- Bug fix https://github.com/kube-burner/kube-burner-ocp/issues/203

### DIFF
--- a/cmd/ocp.go
+++ b/cmd/ocp.go
@@ -93,8 +93,13 @@ func openShiftCmd() *cobra.Command {
 		} else {
 			envVars["ALERTS"] = ""
 		}
-		// If metricsEndpoint is not set, use values from flags
-		if workloadConfig.MetricsEndpoint == "" && esServer != "" && esIndex != "" {
+		esServerEnv := os.Getenv("ES_SERVER")
+		esIndexEnv := os.Getenv("ES_INDEX")
+		if workloadConfig.MetricsEndpoint == "" && (esServerEnv == "" || esIndexEnv == "") {
+			envVars["ES_SERVER"] = ""
+			envVars["ES_INDEX"] = ""
+		}
+		if esServer != "" && esIndex != "" {
 			envVars["ES_SERVER"] = esServer
 			envVars["ES_INDEX"] = esIndex
 		}


cmd/ocp.go
```diff
@@ -93,6 +93,8 @@ func openShiftCmd() *cobra.Command {
 		} else {
 			envVars["ALERTS"] = ""
 		}
+                envVars["ES_SERVER"] = ""
+                envVars["ES_INDEX"] = ""
```
krishvoor, line 97: Formatting nit, can you run a `gofmt -s -w cmd/ocp.go`?

krishvoor (approved):
/lgtm

cmd/ocp.go
```diff
@@ -93,6 +93,8 @@ func openShiftCmd() *cobra.Command {
 		} else {
 			envVars["ALERTS"] = ""
 		}
+		envVars["ES_SERVER"] = ""
```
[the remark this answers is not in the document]
vishnuchalla: We need to make sure the it fails on ES_SERVER when a `--metrics-endpoint` is specified which is supposed to have es values provided.

cmd/ocp.go
```diff
@@ -93,6 +93,17 @@ func openShiftCmd() *cobra.Command {
...
+		esIndexEnv := os.Getenv("ES_INDEX")
+		if workloadConfig.MetricsEndpoint == "" && (esServerEnv == "" || esIndexEnv == "") {
+			envVars["ES_SERVER"] = ""
+			envVars["ES_INDEX"] = ""
+		}
+		if esServer != "" && esIndex != "" {
+			envVars["ES_SERVER"] = esServer
+			envVars["ES_INDEX"] = esIndex
+		}
+		envVars["ES_INDEX"] = ""
```
vishnuchalla, line 106:
```suggestion
		envVars["ES_INDEX"] = ""
```
This specific line should be removed as it overrides and legit ES_INDEX value to empty string.

rsevilla87:
Im a bit lost regarding this issue, after digging a bit, I saw that line 97 was updated by https://github.com/kube-burner/kube-burner-ocp/pull/172, we could fix this issue more easily by putting the original conditional back.

I'm probably missing some context on that change due to my inactivity period, am I missing something @vishnuchalla ?

cmd/ocp.go
```diff
@@ -93,6 +93,17 @@ func openShiftCmd() *cobra.Command {
...
+		esServerEnv := os.Getenv("ES_SERVER")
+		esIndexEnv := os.Getenv("ES_INDEX")
+		if workloadConfig.MetricsEndpoint == "" && (esServerEnv == "" || esIndexEnv == "") {
+			envVars["ES_SERVER"] = ""
+			envVars["ES_INDEX"] = ""
+		}
+		if esServer != "" && esIndex != "" {
+			envVars["ES_SERVER"] = esServer
+			envVars["ES_INDEX"] = esIndex
+		}
```
rsevilla87, line 105: We shouldn't use external env vars

```suggestion
		if workloadConfig.MetricsEndpoint ==  "" && (esServer != "" && esIndex != "") {
			envVars["ES_SERVER"] = esServer
			envVars["ES_INDEX"] = esIndex
		} else{
			envVars["ES_SERVER"] = ""
			envVars["ES_INDEX"] = ""
		}
		
```

vishnuchalla: This won't work when `MetricsEndpoint` is specified and has a `${ES_SERVER}` variable which is exported as an environment variable (.i.e in case of our CI tests), I beleive.

smalleni: I'll take a look at this later this week, yeh' I didn't think of the metrics-endpoint either and didn't factor that into my approach.

smalleni: `envVars["ES_SERVER"] = esServer` was all that was needed I thought, if you look at my initial patch. Let's hold this and I'll revisit this later this week. This is not a priority for me right now.

smalleni: Vishnu is right, we need to account for cases where metrics endpoitn is used and consumes variables from the env.

smalleni:
@vishnuchalla @rsevilla87 Vishnu's concern is valid when using metricsendpoint and using env vars within that. So I just updated the commit with his suggestion.

afcollins (approved):
LGTM

cmd/ocp.go
rsevilla87: Exporting the ES_SERVER and ES_INDEX env variables is a CPT/e2-benchmarking thing, these variables are not documented in kube-burner-ocp and their values are actually passed as flags `--es-server` and `--es-index` respectively

rsevilla87: We shouldn't read those variables from kube-burner-ocp

rsevilla87: This is not correct, we should use the flags

```
$ ES_SERVER="blabla" ES_INDEX=foo ./bin/amd64/kube-burner-ocp cluster-density-v2 --iterations=1  
time="2025-04-10 11:22:07" level=info msg="❤️  Checking for Cluster Health" file="cluster_health.go:46"
time="2025-04-10 11:22:09" level=info msg="Cluster is Healthy" file="cluster_health.go:54"
time="2025-04-10 11:22:11" level=info msg="📁 Creating opensearch indexer: indexer-0" file="metrics.go:67"
time="2025-04-10 11:22:11" level=fatal msg="Error creating indexer 0: OpenSearch health check failed: unsupported protocol scheme \"\"" file="metrics.go:70"
```

PR #211, opened 2025-03-31
add cnv related metrics

gqlo:
## Type of change

- New feature

## Description

Add cnv related metrics

### DIFF
--- a/cmd/config/metrics-report.yml
+++ b/cmd/config/metrics-report.yml
@@ -225,7 +225,40 @@
   instant: true
 
 # OpenShift virtualization
+## metadata
+- query: kubevirt_hyperconverged_operator_health_status{}
+  metricName: hco-operator-health-cnv
+  instant: true
 
+- query: sum(kubevirt_number_of_vms)
+  metricName: total-vm-created-cnv
+  instant: true
+
+- query: sum(cnv:vmi_status_running:count)
+  metricName: total-vmi-running-cnv
+  instant: true
+
+- query: cluster:vmi_request_cpu_cores:sum
+  metricName: total-cpu-request-cnv
+  instant: true
+
+- query: sum(kubevirt_vm_resource_requests)
+  metricName: total-mem-request-cnv
+  instant: true
+
+- query: kubevirt_virt_api_up
+  metricName: virt-api-pod-count-cnv
+  instant: true
+
+- query: sum(kubevirt_memory_delta_from_requested_bytes)
+  metricName: control-plane-pod-mem-usage-request-delta-cnv
+  instant: true
+
+- query: sum(kubevirt_vmi_launcher_memory_overhead_bytes{}) / count (kubevirt_vmi_launcher_memory_overhead_bytes{})
+  metricName: avg-virt-launcher-mem-overhead
+  instant: true
+
+## CPU and Memory related metircs
 - query: avg(avg_over_time(irate(container_cpu_usage_seconds_total{name!="", namespace="openshift-cnv", container!="POD"}[2m])[{{.elapsed}}:])) by (container)
   metricName: cpu-cnv
   instant: true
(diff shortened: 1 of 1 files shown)


josecastillolema:
@gqlo can you please fix the DCO faling test?

josecastillolema:
@jeniferh @jtaleric @rsevilla87 do you think we should keep this metrics in the generic metrics-report file or create a new file specific for CNV?

josecastillolema:
Have you tested this file with kube-burner-ocp @gqlo ? Can you share an output of the testing?

gqlo:
> @gqlo can you please fix the DCO faling test?

Hi Jose, fixed the doc check

vishnuchalla (approved):
lgtm

vishnuchalla:
@rsevilla87 can you take a look as well? Seems like good to go.

jeniferh:
Agreed its probably best if we can separate to collect for virt-only runs, otherwise its just potentially confusing noise.

rsevilla87:
@gqlo do you have an example of how these metrics look like?, you're adding them to the "report" metrics profile which is a profile meant to return only a few datapoints (normally 1 per query) and I fear that some of them doesn't fit here.

i.e: 
- control-plane-pod-cpu-usage-cnv
- vmi-read-iops-cnv
- vmi-write-iops-cnv
- vmi-storage-write-traffic-cnv

As mentioned previously, we should move these metrics to a different profile and make sure of what do we want to scrape and how to consume it. (some of them are redundant IMO)

rsevilla87:
Also, it's important to note that kube-burner-ocp supports using multiple metrics-profiles.i.e:

```shell
$ kube-burner-ocp kubevirt-density --metrics-profile metrics-aggregated.yml,metrics-virtualization.yml
```

josecastillolema:
> Also, it's important to note that kube-burner-ocp supports using multiple metrics-profiles.i.e:
> 
> ```shell
> $ kube-burner-ocp kubevirt-density --metrics-profile metrics-aggregated.yml,metrics-virtualization.yml
> ```

Oh I didn't know that, that is great that way we avoid duplicating metrics on several files and we can reports focused on specific functionality (i.e.: CNV)

jtaleric:
yeah - i think we just move these to `cnv-metrics.yml` -- wdyt?

gqlo:
> @gqlo do you have an example of how these metrics look like?, you're adding them to the "report" metrics profile which is a profile meant to return only a few datapoints (normally 1 per query) and I fear that some of them doesn't fit here.
> 
> i.e:
> 
> * control-plane-pod-cpu-usage-cnv
> * vmi-read-iops-cnv
> * vmi-write-iops-cnv
> * vmi-storage-write-traffic-cnv
> 
> As mentioned previously, we should move these metrics to a different profile and make sure of what do we want to scrape and how to consume it. (some of them are redundant IMO)
@rsevilla87  Hi Raul it looks like this. 
Screenshot from 2025-05-15 09-36-22

gqlo:
> yeah - i think we just move these to `cnv-metrics.yml` -- wdyt?

yeah makes sense, I can move thest to `cnv-metrics.yml`

> yeah - i think we just move these to `cnv-metrics.yml` -- wdyt?

yeah makes sense, i can move it to cnv-metrics.yml, this should be under the same folder, right?

rsevilla87:
> > yeah - i think we just move these to `cnv-metrics.yml` -- wdyt?
> 
> @jtaleric yeah makes sense, i can move it to cnv-metrics.yml, this should be under the same folder, right?

correct

gqlo:
> > > yeah - i think we just move these to `cnv-metrics.yml` -- wdyt?
> > 
> > 
> > @jtaleric yeah makes sense, i can move it to cnv-metrics.yml, this should be under the same folder, right?
> 
> correct

@rsevilla87  Just moved all cnv realted metrics to `cnv-metrics.yml`

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,119 @@
...
+- query: cluster:vmi_request_cpu_cores:sum
+  metricName: total-cpu-request-cnv
+  instant: true
+
+- query: sum(kubevirt_vm_resource_requests)
+  metricName: total-mem-request-cnv
+  instant: true
+
+- query: kubevirt_virt_api_up
+  metricName: virt-api-pod-count-cnv
```
josecastillolema, line 25: Is this name correct? Seems like a typo

gqlo: this query tracks the number of virt-api pod running.

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,119 @@
+---
+# OpenShift virtualization
+## metadata
+- query: kubevirt_hyperconverged_operator_health_status{}
```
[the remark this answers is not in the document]
gqlo: Can you point me to where is that alter implemented?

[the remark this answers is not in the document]
josecastillolema: https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/alerts.yml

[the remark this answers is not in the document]
gqlo: I could move some of those metrics there,  what if users are using kube-burner in a env where CNV operators are not installed, and those metrics won't be available, I hope kube-burner is hanlding such exceptions.

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,124 @@
+---
+# OpenShift virtualization
+# elapsed is the job duration time
+## metadata
+- query: kubevirt_hyperconverged_operator_health_status{}
+  metricName: hco-operator-health
+  instant: true
```
rsevilla87, line 7: I think it could be a good practice to fetch this metric without instant, in order to track HCO's health during the whole workload runtime, another alternative is to add a new alerts-cnv.yml file to track KPIs like this one, find more help about alerts here:

https://kube-burner.github.io/kube-burner/latest/observability/alerting/

and more examples at:

https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/alerts-profiles/alerts.yml

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,124 @@
...
+
+- query: cluster:vmi_request_cpu_cores:sum
+  metricName: total-cpu-request
+  instant: true
+
+- query: sum(kubevirt_vm_resource_requests)
+  metricName: total-mem-request
+  instant: true
+
+- query: sum(up{namespace="openshift-cnv",pod=~"virt-api-.*"}) or vector(0)
```
rsevilla87, line 25: What's the meaning of this expression?

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,124 @@
+---
+# OpenShift virtualization
+# elapsed is the job duration time
+## metadata
+- query: kubevirt_hyperconverged_operator_health_status{}
+  metricName: hco-operator-health
+  instant: true
+
+- query: sum(kubevirt_number_of_vms)
```
rsevilla87, line 9: This two metrics can be pulled without instant: true

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,124 @@
...
+  instant: true
+
+## network related metrics
+- query: avg(avg_over_time(rate(kubevirt_vmi_network_receive_bytes_total[60s])[{{.elapsed}}:])) by (name)
+  metricName: avg-vmi-traffic-received
+  instant: true
+
+- query: avg(avg_over_time(rate(kubevirt_vmi_network_transmit_bytes_total[60s])[{{.elapsed}}:])) by (name) 
+  metricName: avg-vmi-traffic-transmitted
+  instant: true
```
rsevilla87, line 94: network usage metrics are not relevant with the tests we have at the moment

cmd/config/cnv-metrics.yml
```diff
@@ -0,0 +1,124 @@
...
+  metricName: migration-in-scheduling-phase
+  instant: true
+
+- query: sum(kubevirt_vmi_migrations_in_running_phase)
+  metricName: migration-in-running-phase
+  instant: true
+
+- query: sum(kubevirt_vmi_migrations_in_pending_phase)
+  metricName: migration-in-pending-phase
+  instant: true
```
rsevilla87, line 124: I think we should avoid migration metrics for the moment, as they are only relevant in migration in scenarios running migrations, which is not the case. @ygalblum has been working on some of them, but I think he's not currently using kube-burner's indexing capabilities.


Also, if we are adding these metrics, they shouldn't be instant, as migrations is an operation that happens during the whole workoad runtime and pulling the last datapoint won't probably return anything relevant

cmd/config/cnv-metrics.yml
gqlo: This is getting the number of running virt-api pods, if none exists, returns 0.

cmd/config/cnv-metrics.yml
gqlo: what about removing the instant and keep them for now? At some point, we will add some migration tests.

cmd/config/cnv-metrics.yml
gqlo: I suggest keeping those metrics for now. As charles is working on the UDN tests, we might need it soon.

qktrn:
@rsevilla87 removed instant label.

gqlo:
CC @rsevilla87

rsevilla87 (approved):
lgtm

PR #212, opened 2025-03-31
Virt Clone - remove restart steps

ygalblum:
## Type of change

- Refactor
- Optimization

## Description

Remove the VM restart steps from this test as they will be moved to a new dedicated test
Simplify the YAML file


vishnuchalla:
CI tests are running into virtctl failures.

PR #213, opened 2025-03-31
UDN Churn fix

smalleni:
The way that the udn-density-pods workload has currently been implemented, the churn functionality does not ireally work as the UDNs are created in a separate job from the one in which pods/other resoruces are created. This leads to a condition where the `udn-density-l3-pods` job tries to create pods without the UDN, which has been deleted during the churn phase. Thecurrent changes preserve the jobPause functionality by keeping UDN creationin a separate job and enable true churning without pause when requested by user. Churn is also defaulted to false after this change. Current behavior with jobPause is not changed with this patch.

## Type of change

- Refactor
- Bug fix


chentex (approved):
/lgtm

mohit-sheth (approved):
lgtm
tested with 72 iterations 
UUID 93a95693-986d-4d09-b5e7-6bf04594cf6d on AWS Pro instance

PR #217, opened 2025-04-07
Remove web-burner tests

rsevilla87:
## Type of change

- Refactor
- Documentation
- CI

## Description

Let's lose some weight here.

- Related Issue #
- Closes #


josecastillolema:
@rsevilla87 any particular reason to do it now?
I mean it is officially deprecated for now, but what if at some point in the near future some customer comes with demands for testing this?

If it is not bothering other stuff, I would leave this workload for now in the repo (marked as deprecated), just to be on the safe side.

What do you think @Noreen21 ?

vishnuchalla:
> @rsevilla87 any particular reason to do it now? I mean it is officially deprecated for now, but what if at some point in the near future some customer comes with demands for testing this?
> 
> If it is not bothering other stuff, I would leave this workload for now in the repo (marked as deprecated), just to be on the safe side.
> 
> What do you think @Noreen21 ?

I think, if its dead code we can get it removed. If we want to fetch it back it should always remain in github commit history. Or @josecastillolema you can create you own triage repo where you can store stuff like this. wdyt?

rsevilla87:
> > @rsevilla87 any particular reason to do it now? I mean it is officially deprecated for now, but what if at some point in the near future some customer comes with demands for testing this?
> > If it is not bothering other stuff, I would leave this workload for now in the repo (marked as deprecated), just to be on the safe side.
> > What do you think @Noreen21 ?
> 
> I think, if its dead code we can get it removed. If we want to fetch it back it should always remain in github commit history. Or @josecastillolema you can create your own triage repo where you can store stuff like this. wdyt?

As per the conversation I held with @josecastillolema, I've updated the PR to only remove web-burner tests from CI.
Seems like the telco team cannot confirm if web-burner will be necessary in the future so let's keep it around for some more time

Noreen21:
@rsevilla87 @josecastillolema @vishnuchalla +1 on removing the web-burner tests from CI. As long as we have a backup somewhere in case the code is needed again, it shouldn't matter how we are storing it - be it in a separate triage repo or depracating (yet keeping) it in the same repo. We can revisit this once we have an established Core RDS CI.

PR #219, opened 2025-04-08
Consolidate node-density workloads into a single file

rsevilla87:
## Type of change

- Refactor

## Description

Remove some duped stuff by consolidating node-density workloads into a single file


vishnuchalla (approved):
lgtm

cmd/config/node-density-heavy/node-density-heavy.yml
```diff
@@ -26,6 +26,10 @@ global:
         labelSelector: {app: ovnkube-control-plane}
         url: http://localhost:29108/debug/pprof/profile?seconds=30
 {{ end }}
+{{ if eq .SVC_LATENCY "true" }}
```
ygalblum, line 29: Not a must for now. But, you can avoid having to transform the variables to and from environment variables by passing a `key:value` map to the `WorkloadHelper` by calling `RunWithAdditionalVars` (https://github.com/kube-burner/kube-burner/blob/93feb51810e04cb13c0a23b30912bbfc27e60344/pkg/workloads/helpers.go#L52) instead of `Run`

PR #220, opened 2025-04-08
add new measurements support

venkataanil:
This enables kube-burner-ocp to define its own measurement functionality (example, network-policy-measurement.go, router-advertisement-measurement.go). This is required to avoid using openshift CRD in kube-burner measurements code.

It defines exampleLatency which is currently working and can be used as a reference for other measurements like router-advertisement-measurement.go.

Now kube-burner-ocp prepares measurementFactoryMap with the core measurements (like pod, service latency) and additional measurements which is defined in kube-burner-ocp (like exampleLatency).

I have tested this code with only enabling podLatency and exampleLatency. I will fine tune it once we are fine with this approach.

This PR depends on kube-burner PR
https://github.com/kube-burner/kube-burner/pull/841

## Type of change

- Refactor
- New feature
- Bug fix
- Optimization
- Documentation
- CI

- Related Issue #
- Closes #

### DIFF
--- a/cluster-density.go
+++ b/cluster-density.go
@@ -20,11 +20,16 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/kube-burner/kube-burner/pkg/config"
+	"github.com/kube-burner/kube-burner/pkg/measurements"
 	"github.com/kube-burner/kube-burner/pkg/workloads"
 	log "github.com/sirupsen/logrus"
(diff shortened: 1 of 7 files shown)


venkataanil:
This PR is opened to show how we can add measurements in kube-burner-ocp to support https://github.com/kube-burner/kube-burner/pull/841

PR #226, opened 2025-04-16
CI - Restore tmp in PATH to allow access to virtctl

ygalblum:
## Type of change

- Bug fix
- CI

## Description

virtctl is downloaded to /tmp so it must be in the PATH


vishnuchalla (approved):
lgtm

PR #234, opened 2025-04-29
Move ralatency to pkg and implement only for Linux

### LINKED ISSUE #233: [BUG] Route Advertisement Latency breaks compilation on non Linux OS
# Bug Description
See Action: https://github.com/kube-burner/kube-burner-ocp/actions/runs/14732474661

ygalblum:
## Type of change

- Bug fix

## Description

The Route Advertisement Latency measurement has Linux dependencies. As a result, it fails the build for other OSs. Move to a new package and separate the implementation based on the OS  

- Closes #233


vishnuchalla (approved):
lgtm

PR #243, opened 2025-05-08
Fix logic to sort tags based on creation date

vishnuchalla:
## Type of change

- Bug fix
- Documentation
- CI

## Description
Previous logic to fetch all the tags used to sort them in descending order lexicographically, which wasn't what we wanted in reality. So fixing that logic to fetch the logs and sort them solely based on creation times so that we fetch and keep only 3 latest version of them and delete the rest.

There have been failures in deploying docs in the past due to this issue.

## Testing
Tested and verified these changes locally in a hacky way: https://github.com/kube-burner/kube-burner/pull/858


rsevilla87 (approved):
lgtm

PR #253, opened 2025-05-24
use RA API v2

venkataanil:
routeadvertisement API changed which introduced new fields in networkSelectors. Instead of using ovnk's UDN and RA clientsets, we are now using dynamic client and informers with GVR to watch and fetch UDN and RA.

We will use metrics.yml instead of metrics-aggregated.yml, similar to other UDN workload.

CUDN will be using 40.X.X.X CIDR instead of 10.X.X.X to make BGP route exchange work with AWS cluster.


go.sum
```diff
@@ -54,6 +54,8 @@ github.com/chzyer/readline v1.5.1/go.mod h1:Eh+b79XXUwfKfcPLepksvw2tcLE/Ct21YObk
 github.com/chzyer/test v0.0.0-20180213035817-a1ea475d72b1/go.mod h1:Q3SI9o4m/ZMnBNeIyt5eFwwo7qiLfzFZmjNmxjkiQlU=
 github.com/chzyer/test v1.0.0/go.mod h1:2JlltgoNkt4TW/z9V/IzDdFaMTM2JPIi26O1pF38GC8=
 github.com/client9/misspell v0.3.4/go.mod h1:qj6jICC3Q7zFZvVWo7KLAzC3yx5G7kyvSDkc90ppPyw=
+github.com/cloud-bulldozer/go-commons v1.0.19 h1:WVkKb5uTyxFmrsRqZKU4RT7BBhjystpdSgm7VPpJSlw=
+github.com/cloud-bulldozer/go-commons v1.0.19/go.mod h1:CKmdWhyxN6p4leLhdwKxQHJ3wiKNUistFKsctLTSqIk=
```
rsevilla87, line 58: We shouldn't import go-commons v1, PTAL

vishnuchalla:
Except for go-commons version everything looks good to me.

rsevilla87 (approved):
lgtm

PR #262, opened 2025-06-10
Adding metricsClosing

rsevilla87:
## Type of change

- Bug fix

## Description

Due to https://github.com/kube-burner/kube-burner/pull/852, metrics have been affected since metrics are not collected during the pause period by default.

Also removing unnecessary jobPause flag (which by default was configured to 1 ms), and last skip indexing of the first workload stage

- Related Issue https://github.com/kube-burner/kube-burner/issues/886
- Closes #

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -1,3 +1,3 @@
 [codespell]
 skip = go.mod,go.sum
-ignore-words-list = passt
+ignore-words-list = passt, NotIn
--- a/README.md
+++ b/README.md
@@ -444,6 +444,9 @@ By default, the test will search for the `StorageClass` to use:
 
 To use a different one, use `--storage-class` to provide a different name.
 
+In addition, multiple `StorageClasses` can be used by passing a comma separated list names.
+The test will then choose a different `StorageClass` for each loop in round robin.
+
 Please note that regardless to which `StorageClass` is used, it must:
 - Support Volume Expansion: `allowVolumeExpansion: true`.
 - Have a corresponding `VolumeSnapshotClass` using the same provisioner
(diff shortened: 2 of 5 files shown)


cmd/config/udn-density-pods/udn-density-pods.yml
```diff
@@ -60,7 +60,7 @@ jobs:
     waitWhenFinished: true
     preLoadImages: false
     preLoadPeriod: 15s
-    jobPause: {{.JOB_PAUSE}}
+    skipIndexing: true
```
vishnuchalla, line 63: One the other hand, why do we want to skip the indexing?

udn-density-pods.go
```diff
@@ -69,7 +69,7 @@ func NewUDNDensityPods(wh *workloads.WorkloadHelper) *cobra.Command {
 		},
 	}
 	cmd.Flags().BoolVar(&l3, "layer3", true, "Layer3 UDN test")
-	cmd.Flags().StringVar(&jobPause, "job-pause", "1ms", "Time to pause after finishing the job")
+	cmd.Flags().DurationVar(&jobPause, "job-pause", 0, "Time to pause after finishing the job that creates the UDN")
```
venkataanil, line 72: This may impact prow runs as we are passing string --job-pause=45m in CI
https://github.com/openshift/release/blob/master/ci-operator/config/openshift-eng/ocp-qe-perfscale-ci/openshift-eng-ocp-qe-perfscale-ci-main__aws-4.19-nightly-x86.yaml#L393

venkataanil (commented):
Lets only retain  "metricsClosing: afterJobPause" change and avoid other optimizations as this PR is to address the CI regression.

venkataanil (approved):
LGTM

PR #265, opened 2025-06-17
Update kube-burner import tag

ygalblum:
## Type of change

- Bug fix

## Description

Update kube-burner import tag to fix reading template for patch


vishnuchalla (approved):
lgtm

PR #270, opened 2025-06-25
Updating CDv2 pod ready latency probe

mukrishn:
## Type of change

- [ ] Refactor
- [ ] New feature
- [ ] Bug fix
- [x] Optimization
- [ ] Documentation
- [ ] CI

## Description
The current CDv2 workload use of `readinessProbe` is overwhelming the OpenShift-Ingress `router` in very large-scale clusters (> 250 nodes), due to thousands of routes being probed by client pods every 10s. This causes the router’s CPU usage to spike over 5 cores, and it also impacts pod startup latency, as readiness checks from newly progressing pods compete with ongoing traffic from existing client pods.
AWS handles this scenario well while Azure is struggling at 250-node scale, primarily due to its weaker ingress performance. With 500 node it gets even worse, 9K routes with ~22K client pods at a 10s interval.

In this change,  updating the behavior by replacing the `readinessProbe` with a `startupProbe` that would run only once and mark the container `ready`. This evaluates container readiness latency based on the first successful response(may be we can increase success threshold) from service and route endpoints. This change allows the container to be marked as `Ready` immediately after initial verification as we don't require to monitor health of the pod continuously every 10s.

Additionally, to maintain active traffic between pods/svcs, a separate curl job runs in the container at a reduced frequency (every 2 minutes), minimizing impact on the router while still validating end-to-end connectivity and logging. Having another `readinessProbe` with lower interval instead of this job may still affect the latency at large scale. 

Client logs
```
2025-06-25 17:05:36: INFO curl on http://cluster-density-4/256.html and https://cluster-density-1-cluster-density-v2-1.apps.rhperfscale.org/256.html
curl: (22) The requested URL returned error: 503
2025-06-25 17:07:36: INFO curl on http://cluster-density-4/256.html and https://cluster-density-1-cluster-density-v2-1.apps.rhperfscale.org/256.html
2025-06-25 17:09:36: INFO curl on http://cluster-density-4/256.html and https://cluster-density-1-cluster-density-v2-1.apps.rhperfscale.org/256.html
2025-06-25 17:11:37: INFO curl on http://cluster-density-4/256.html and https://cluster-density-1-cluster-density-v2-1.apps.rhperfscale.org/256.html
```

- Related Issue #
- Closes #

### DIFF
--- a/cmd/config/cluster-density-v2/deployment-client.yml
+++ b/cmd/config/cluster-density-v2/deployment-client.yml
@@ -34,7 +34,10 @@ spec:
       containers:
       - name: client-app
         image: quay.io/cloud-bulldozer/curl:latest
-        command: ["sleep", "inf"]
+        command:
+            - "/bin/sh"
+            - "-c"
+            - "while true; do echo $(date +'%Y-%m-%d %H:%M:%S'): INFO curl on ${ROUTE_ENDPOINT}; curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null; sleep 120; done"
         resources:
           requests:
             memory: "10Mi"
@@ -47,7 +50,7 @@ spec:
             command:
             - "/bin/sh"
             - "-c"
-            - "curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null && curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null"
+            - "if [ ! -f /tmp/ready.log ]; then curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null && curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null && touch /tmp/ready.log; else curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null; fi"
           periodSeconds: 10
           timeoutSeconds: 5
           failureThreshold: 3
--- a/cmd/config/cluster-density-v2/deployment-client.yml
+++ b/cmd/config/cluster-density-v2/deployment-client.yml
@@ -37,7 +37,7 @@ spec:
         command:
             - "/bin/sh"
             - "-c"
-            - "while true; do echo $(date +'%Y-%m-%d %H:%M:%S'): INFO curl on ${ROUTE_ENDPOINT}; curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null; sleep 120; done"
+            - "while true; do echo curl on ${ROUTE_ENDPOINT}; curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null; sleep 120; done"
         resources:
           requests:
             memory: "10Mi"


afcollins:
Interesting proposal! What sort of impact does this have on podReady latency distribution?

cmd/config/cluster-density-v2/deployment-client.yml
```diff
@@ -34,15 +34,18 @@ spec:
       containers:
       - name: client-app
         image: quay.io/cloud-bulldozer/curl:latest
-        command: ["sleep", "inf"]
+        command:
+            - "/bin/sh"
+            - "-c"
+            - "while true; do echo $(date +'%Y-%m-%d %H:%M:%S'): INFO curl on ${SERVICE_ENDPOINT} and ${ROUTE_ENDPOINT}; curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null && curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null; sleep 120; done"
```
rsevilla87, line 40: Using commands is not the same as probes, as they don't generate load in the kubelet/cri-o side

rsevilla87: Im not sure how the pod will behave when a command like this fails, it will probably continue its execution regarldes if there's an error or not.

mukrishn: > they don't generate load in the kubelet/cri-o side
True, this looping curl is simply to generate a live ingress traffic in the cluster. If you suggest to keep some active probes, can we add a probe for the sake of kubelet/cri-o may a probe that always gets true ? something that is not overwhelming router.

mukrishn: Or I can do a conditional readiness probe, that would curl routes once and be simpler later that runs every 10s. 
```sh
            - "if [ ! -f /tmp/ready.log ]; then curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null && \ 
               curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null && touch /tmp/ready.log; \ 
               else curl --fail -sS ${SERVICE_ENDPOINT} -o /dev/null; fi" 
```

vishnuchalla: This command is fine as a readiness probe, where we just want to see if we are ready to serve the traffic in the beginning but for the rest of the period while do a 
```
            - "while true; do echo curl on ${ROUTE_ENDPOINT}; curl --fail -sSk ${ROUTE_ENDPOINT} -o /dev/null; sleep 120; done"
```
which basically is to see, if my app is still up and running, liveness probe would be the best fit instead of a command.

vishnuchalla: How about setting `failureThreshold` high enough to a value that actually can be considered as a failure?

mukrishn: Agree we can do, but additional probes would increase the load on kubelet. Anything is fine as long as we are not changing the load. As I said our intention of this command is only to generate some traffic from the client pod against the route endpoint without any load.

vishnuchalla (approved):
lgtm

mukrishn:
CDv2 on 24 node ROSA - [Link](https://grafana.rdu2.scalelab.redhat.com:3000/d/D5E8c5XVz/kube-burner-report-mode?orgId=1&from=1751328000000&to=1751500799000&var-Datasource=ab3f14e6-a50f-4d52-93fa-a5076794f864&var-platform=AWS&var-sdn=OVNKubernetes&var-clusterType=All&var-job=cluster-density-v2&var-workerNodesCount=24&var-ocpMajorVersion=4.19&var-uuid=6142ba97-6add-4a2e-86e6-1b68d3f14929&var-uuid=a23dc112-35c1-428e-8450-a6320786212c&var-compare_by=uuid&var-component=kube-apiserver&var-component=kube-controller-manager&var-node_roles=masters&var-node_roles=workers&var-node_roles=infra), Perf looks similar across except slight deviation in the API latency, I guess it is not related. Let me know if further tests required.

`a23dc112-35c1-428e-8450-a6320786212c` - with changes
`6142ba97-6add-4a2e-86e6-1b68d3f14929` - default cdv2

PR #275, opened 2025-07-02
Replace Run() by RunWithAdditionalVars()

### LINKED ISSUE #238: [RFE] Avoid having to transform the variables to and from env vars by calling `RunWithAdditionalVars`
Not a must for now. But, you can avoid having to transform the variables to and from environment variables by passing a `key:value` map to the `WorkloadHelper` by calling `RunWithAdditionalVars` (https://github.com/kube-burner/kube-burner/blob/93feb51810e04cb13c0a23b30912bbfc27e60344/pkg/workloads/helpers.go#L52) instead of `Run`

_Originally posted by @ygalblum in https://github.com/kube-burner/kube-burner-ocp/pull/219#discussion_r2054060118_

rsevilla87:
## Type of change

- Optimization

## Description

Using RunWithAdditionalVars saves us to deal with string conversions, and pollute the environment vars

- Related Issue #
- Closes #238


whereabouts.go
```diff
@@ -37,14 +36,17 @@ func NewWhereabouts(wh *workloads.WorkloadHelper) *cobra.Command {
 		Short:        "Runs whereabouts workload",
 		SilenceUsage: true,
 		PreRun: func(cmd *cobra.Command, args []string) {
-			os.Setenv("JOB_ITERATIONS", fmt.Sprint(iterations))
-			os.Setenv("POD_READY_THRESHOLD", fmt.Sprintf("%v", podReadyThreshold))
-			os.Setenv("CONTAINER_IMAGE", containerImage)
-			os.Setenv("FAST", fmt.Sprint(fast))
+			additionalVars := map[string]any{
```
ygalblum, line 39: Not sure what you did here. You're running the workload twice (in `PerRun` and in `Run` with and without the parameters

ygalblum (commented):
General comment about the CI failures. This change also changes the variable types. Environment variables are all interpreted as strings. But, once you use the additional vars  map, they can be of different types. For example `PPROF` used to be a string and now its a boolean

whereabouts.go
vishnuchalla: Looks like `PreRun` has been removed and this is in the actual run now.

rsevilla87:
> Yes, the right thing to do is to adjust the templates. But, I think you can use booleans as booleans without `eq`. I gave two examples (for `true` and for `false`), but they are for all cases

Thanks!, to be honest, I didn't know that was possible, I always find go template documentation confusing.

rsevilla87:
> I think you added a results directory by mistake (`collected-metrics-abcd`)

yeah, the change become more complex that what I initially thought :)

egressip.go
```diff
@@ -154,16 +152,17 @@ func NewEgressIP(wh *workloads.WorkloadHelper, variant string) *cobra.Command {
...
-			os.Setenv("POD_READY_THRESHOLD", fmt.Sprintf("%v", podReadyThreshold))
-			os.Setenv("ADDRESSES_PER_ITERATION", fmt.Sprint(addressesPerIteration))
-			os.Setenv("EXTERNAL_SERVER_IP", externalServerIP)
-			generateEgressIPs(iterations, addressesPerIteration, externalServerIP)
-		},
 		Run: func(cmd *cobra.Command, args []string) {
 			setMetrics(cmd, metricsProfiles)
-			rc = wh.Run(cmd.Name() + ".yml")
+			eipAddresses := strings.Join(generateEgressIPs(iterations, addressesPerIteration, externalServerIP), " ")
+			additionalVars := map[string]any{
```
ygalblum, line 158: Shouldn't you use `AdditionalVars`?

ygalblum (approved):
LGTM

PR #276, opened 2025-07-04
Virt Migration - allow users to set QPS for the migration phase

ygalblum:
## Type of change

- Optimization
- Documentation
- CI

## Description

Some clusters are overloaded when handling 20 parallel requests. Allow users to set a different value


vishnuchalla (approved):
lgtm

PR #281, opened 2025-07-14
check for bool instead of string for SIMPLE in udn test

mohit-sheth:
## Type of change

- Bug fix

declared as a bool https://github.com/kube-burner/kube-burner-ocp/blob/b43c6416dc32a3fd5366a6c0011eb98fc013329a/udn-density-pods.go#L72


ygalblum (commented):
LGTM

rsevilla87 (approved):

PR #287, opened 2025-07-22
Virt Migration - Add VMIM latency

ygalblum:
## Type of change

- New feature

## Description

Add the collection of vmim latency to the virt-migration test


vishnuchalla (approved):
lgtm

PR #288, opened 2025-07-22
Virt Clone - allow creating VMs in iterations

### LINKED ISSUE #286: [RFE] Allow to configure number of iteration and VM count and number of disk per vm for Virt-clone test
  Yes, currently we are running the virt clone test with 400 VMs in one shot, and it generates high concurrency snapshot requests on the same device ID, and we are hitting the maximum snapshot exceed limit issue with the virt-clone test. So if this test allows us to configure the number of VMs with a disk for each VM, then we can run at a lower scale. 

Virt-clone test should allow for the configuration number of iterations and the disk per VM.

@ygalblum, Can we have a configuration for this test that allows us to set the number of devices per VM and set the number of iteration per VM for this test.

Thanks in advance for your assistance.

ygalblum:
## Type of change

- New feature
- Documentation
- CI

## Description

Virt Clone Test - Allow users to create VMs in iterations instead of all in parallel

- Closes #286


vishnuchalla (approved):
lgtm

PR #298, opened 2025-08-12
Udn vms per node

capolrik:
## Type of change

Adding vms-per-node parameter in virt-udn-density job.

- New feature
- Bug fix

## Description

Making the virt-udn-density job more customizable by modifying the VM density on each node. It impact the number of VM-client deployed on each NS with the following calculation:
```
totalVMs := clusterMetadata.WorkerNodesCount * vmsPerNode
			vmsPerUdn := totalVMs/iterations - 1 // -1 because there is always one server vm per udn
```
All the UDN namespace are identical, so if the the final VM per node is exact but it allow the control of VM density along the UDN density..


virt-density-udn.go
```diff
@@ -82,6 +91,7 @@ func NewVirtUDNDensity(wh *workloads.WorkloadHelper) *cobra.Command {
 	cmd.Flags().IntVar(&churnPercent, "churn-percent", 10, "Percentage of job iterations that kube-burner will churn each round")
 	cmd.Flags().StringVar(&churnDeletionStrategy, "churn-deletion-strategy", "default", "Churn deletion strategy to use")
 	cmd.Flags().IntVar(&iterations, "iteration", 1, "iterations")
+	cmd.Flags().IntVar(&vmsPerNode, "vms-per-node", 100, "VMs per node")
```
jtaleric, line 94: should we target 50 for a default?

capolrik:
@jtaleric @vishnuchalla Is it ok to merge this?

vishnuchalla:
Code changes look good. Need to go through the CI.

PR #306, opened 2025-08-29
Remove pprof collection from non-existent ovn-controller endpoint

afcollins:
## Type of change

- Bug fix

## Description

ovn-controller does not expose a pprof endpoint, so we can remove it from the pprof section to reduce confusion.

```
[root@vm00228 ~]# curl http://localhost:29105/debug/pprof/profile?seconds=5
404 page not found
```


jtaleric (approved):
/lgtm

rsevilla87 (approved):

PR #325, opened 2025-10-15
Remove ok-to-test label requirement from build workflow

rsevilla87:
## Type of change

- Optimization

## Description

To merge a PR this test is required to pass. As it only builds and install the kube-burner-ocp, we can consider it safe because it doesn't interact with the CI cluster.

- Related Issue #
- Closes #


ygalblum (approved):
LGTM

BTW I was about to open the same after I had to add the label on a docs only change

PR #326, opened 2025-10-15
Change default virt-density deletionStrategy to gvr

jeniferh:
## Type of change

- Optimization

## Description

Add deletionStrategy logic to virt-density and default to gvr since it tends to behave better at scale than namespace-level deletion.


rsevilla87 (approved):
lgtm!

PR #330, opened 2025-10-18
Parameterize rds-core hugepages size.

afcollins:
## Type of change

- New feature

## Description

hugepages were hard-coded in the rds-core workload. Now we have an argument to allow them to be configured easily and without modifying the configs.


afcollins:
cc: @josecastillolema

PR #331, opened 2025-10-23
Adding churn to virt density

paigerube14:
## Type of change

- Refactor
- [X] New feature
- Bug fix
- Optimization
- Documentation
- CI

## Description
Want to add the ability to churn on virt density workload to create constant work on the cluster (will perform chaos on this later on)

- Related Issue #
- Closes #


paigerube14:
@rsevilla87 @jeniferh I definitely can change the defaults that they are the same as the behavior we have now. Just wanted to add the option to churn if flags are used

rsevilla87:
> @rsevilla87 is iterations-per-namespace=10 ignored if namespacedIterations is false?

correct, it doesn't have effect.

jeniferh:
Ok thanks for confirming! Otherwise defaults are off so looks ok to me, nice option to have.

rsevilla87 (approved):

jeniferh (approved):

PR #342, opened 2025-11-17
Collect ovs-vswitchd from any slice it runs under

afcollins:
## Type of change

- Bug fix

## Description

When a cluster is configured with performanceprofiles, ovs-vswitchd.service (and ovsdb-server.service, not currently collected) are nested under `/ovs.slice` instead of `/system.slice`.

This fixes the metrics profiles to collect from wherever the service's metrics report to.


vishnuchalla (approved):
lgtm

Issue #359, opened 2025-12-10
[BUG]OpenShift Console error when accessing virt-clone/ virt-capacity-benchmark VMs (e.disk is undefined)


saipramodkola:
# Bug Description
VMs created using the virt-clone/virt-capacity-benchmark  workload in kube-burner-ocp fail to open in the OpenShift Virtualization UI with the error:

**Something went wrong
can't access property "name", e.disk is undefined**

## **Output of `kube-burner-ocp` version**

Version: main
Git Commit: 17b92b877079c0add7af14bfa01a1830753805fd
Build Date: 2025-12-10-05:58:44
Go Version: go1.23.4
OS/Arch: linux amd64

## **Describe the bug**

VMs created using the **virt-clone/virt-capacity-benchmark**  workload in kube-burner-ocp fail to open in the OpenShift Virtualization web console. The console shows the error:

**can't access property "name", e.disk is undefined**

<img width="452" height="211" alt="Image" src="https://github.com/user-attachments/assets/8803035a-0efb-43ee-b668-48c7ec0ba6e2" />

The VM itself is created and runs successfully, but the UI crashes when rendering VM details due to a mismatch between defined volumes and disks in the generated VM spec.

## **To Reproduce**

1. Run the default virt-clone/virt-capacity-benchmark workload:

```kube-burner-ocp virt-clone --storage-class=ocs-storagecluster-ceph-rbd-virtualization --iterations=1 --iteration-clones=10 --access-mode=RWO --log-level debug```

```kube-burner-ocp virt-capacity-benchmark --storage-class=ocs-storagecluster-ceph-rbd-virtualization --max-iterations=1 --vms=10 --data-volume-count=1 --skip-resize-job --skip-migration-job --cleanup=true --log-level debug```

2. Open any of the created VMs in the OpenShift Web Console
3. The UI fails to load the VM details and shows:
     
     can't access property "name", e.disk is undefined

## **Expected behavior**

For every volume defined in the VM spec, there should be a corresponding disk entry.
Specifically, cloudinitdisk should be present under domain.devices.disks.

## **Screenshots or output**

<img width="452" height="211" alt="Image" src="https://github.com/user-attachments/assets/b6ef2381-f481-47e3-b6e0-e0303644a096" />

<img width="452" height="203" alt="Image" src="https://github.com/user-attachments/assets/a96ccc08-428c-4cfc-8d7d-49e136d7a133" />

PR #360, opened 2025-12-10
Fix missing cloudinitdisk in VM disk definitions in Virt-clone and Virt-capacity-benchmark

saipramodkola:
## Type of change

- Bug fix

## Description

This change fixes the **virt-clone** and **virt-capacity-benchmark** VM template by adding a missing disk definition for cloudinitdisk under spec.template.spec.domain.devices.disks.

Previously, the template defined a cloudInitNoCloud volume without a corresponding disk entry, which caused the OpenShift Virtualization web console to fail with the error "**e.disk is undefined**" when rendering VM details.

The update ensures a one-to-one mapping between volumes and disks in the generated VM spec, allowing the OpenShift console to load VM details correctly while preserving existing behavior of the workload.

- Related Issue # https://github.com/kube-burner/kube-burner-ocp/issues/359


saipramodkola:
> Thanks for the fix. While at it, can you please do the same for other such cases: https://github.com/kube-burner/kube-burner-ocp/blob/3adfe7bc791b5520bbf661e29cbcbb5117723a08/cmd/config/virt-migration/templates/vm.yml https://github.com/kube-burner/kube-burner-ocp/blob/3adfe7bc791b5520bbf661e29cbcbb5117723a08/cmd/config/virt-ephemeral-restart/templates/vm.yml.
> 
> Also, please update the description to mention it `Closes` the issue. This way it will be automatically closed

@ygalblum  : I have made the changes to both virt-ephemeral-restart and virt-migration .

saipramodkola:
> The code looks OK to me. But, strange enough the tests failed. I'll rerun them and let's see if there's an issue

From the test summary, it looks like 28 tests ran with 0 failures, but 3 tests timed out (total runtime ~7466 seconds), so the failures appear to be timeout-related , Could you please review once more after the rerun and let me know.

ygalblum (approved):
Bypassing the CI to merge as failures are unrelated

PR #361, opened 2025-12-11
feat/add-jobIterationDelay

luckylittle:
## Type of change

- New features

## Description

Added:
  * `--job-iteration-delay` - Time in `Duration` to wait between job iterations (i.e., between each pod creation). Default is `0`. This can be used to add a delay between iterations to reduce load on the cluster. For example: `--job-iteration-delay=1s` will add a 1 second delay between each pod creation.
  * `--namespace-delay` - Time in `Duration` to wait after completing all iterations in a namespace before starting the next namespace. Default is `0`. This delay occurs after filling up a namespace (completing all `iterations-per-namespace` pods) and before starting the next namespace. For example: `--namespace-delay=30s` will wait 30 seconds after completing all pods in a namespace before starting the next namespace.

Flags added to help:

```bash
./kube-burner-ocp node-density --help | grep job-iteration-delay
      --job-iteration-delay duration     Delay between job iterations
```

```bash
./kube-burner-ocp node-density --help | grep namespace-delay
      --namespace-delay duration         Delay after completing all iterations in a namespace before starting the next namespace
```

All 23 workloads now have these flags available.

## Example

```bash
kube-burner-ocp node-density \
  --iterations-per-namespace 100 \
  --job-iteration-delay 1s \
  --namespace-delay 30s \
  --pods-per-node 800
```

This will run node-density workload on the cluster with 1s delay between each pod creation. After 1 minute and 40 seconds, the namespace is filled up with 100 pods, process will wait 30s before starting filling up the next namespace. This will happen 8 times until there are 800 pods in total.

- Customer in APAC requires this sort of test to avoid OVS saturation issue https://issues.redhat.com/browse/OCPBUGS-65910

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,9 +217,11 @@ This workload is meant to fill with pause pods all the worker nodes from the clu
 
 #### Common Flags
 
-The following flag is available for all workloads:
+The following flags are available for all workloads:
 
 - `--job-iteration-delay` - Time in `Duration` to wait between job iterations (i.e., between each pod creation). Default is `0`. This can be used to add a delay between iterations to reduce load on the cluster. For example: `--job-iteration-delay=1s` will add a 1 second delay between each pod creation.
+
+- `--namespace-delay` - Time in `Duration` to wait after completing all iterations in a namespace before starting the next namespace. Default is `0`. This delay occurs after filling up a namespace (completing all `iterations-per-namespace` pods) and before starting the next namespace. For example: `--namespace-delay=30s` will wait 30 seconds after completing all pods in a namespace before starting the next namespace.
 
 ### node-density-cni
 


rsevilla87:
Hey @luckylittle, thanks for the PR. Some comments:

- You're updating 39 files in a single PR, mostly to remove white spaces, that makes it difficult to review
- The flag `namespaceDelay` doesn't exist, you check out [docs site](https://kube-burner.github.io/kube-burner/v1.17.7/reference/configuration/#jobs) to get an idea of the supported parameters.


Another approach (and recommended) to address the load issues, is to reduce the object creation rate (via --qps and --burst)


And last, you're adding two new AdditionalVars to each job, but not referencing these variables in all of them

PR #371, opened 2026-01-12
Implement set vars feature

### LINKED ISSUE #358: [RFE] Add CLI switch to disable preload
Preload step runs on every execution. However, it is truly only required prior to the first execution of a workload on a cluster, at which point all the images are already loaded.

I believe this is a problem for our KPIs, because we can see resource spikes cluster wide with large number of pod deletions, which can happen as a result of preload daemonset deletion. This spike can throw off our resource averages from what the workload itself actually incurs.

Would like a command line argument to allow users to disable preload easily.

**Describe alternatives you've considered**
This can currently be achieved by extracting the workload, disabling preload in the workload config.

**Additional context**
Also worth noting, that the preload mechanism doesn't do anything more than pull the images. It is not enough to "prime" the cluster for optimal pod ready latencies. In order to get the fastest and most reliable latencies, we usually do a run with the smallest number of iterations (e.g. --pods-per-node=50) to prime the cluster for optimal pod latencies. The first run seems to always have a larger variance than any repeat runs.

rsevilla87:
## Type of change

- New feature

## Description

Implementing the `--set` flag for kube-burner-ocp so we can set arbitrary configuration fields without having to extract and edit the configuration files.

Still hitting some issues due to https://github.com/kube-burner/kube-burner/pull/1072. which should be merged soon.

- Related Issue #
- Closes #358


ygalblum (approved):
LGTM

PR #379, opened 2026-01-31
Netpol migration

rsevilla87:
## Type of change

- Refactor

## Description

Moving the network policy measurement here, this measurement does not follow kube-burner's standards, and can be misleading to have it in the core repository.

cc @venkataanil 

- Related Issue #
- Closes #

### DIFF
--- a/cmd/config/network-policy/network-policy.yml
+++ b/cmd/config/network-policy/network-policy.yml
@@ -2,14 +2,6 @@
 global:
   gc: true
   gcMetrics: false
-  measurements:
-{{ if .NETPOL_LATENCY }}
-    - name: netpolLatency-test
-      thresholds:
-        - conditionType: Ready
-          metric: P99
-          threshold: {{.NETPOL_READY_THRESHOLD}}
-{{ end }}        
 metricsEndpoints:
 {{ if .ES_SERVER }}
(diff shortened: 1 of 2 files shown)


cmd/config/network-policy/network-policy.yml
```diff
@@ -58,6 +50,14 @@ jobs:
         replicas: 1
 
   - name: network-policy-perf
+  {{ if .NETPOL_LATENCY }}
```
rsevilla87: @venkataanil, now it's possible to specify measurements per job, so the hardcoded skip logic is not required anymore.

rsevilla87:
> Please let me know what kube-burner's standards this measurement is not following so that we can improve it.

as we discussed, the measurement is not standard, and doesn't work OOTB, it needs some hardcoded tricks that can be confusing, Im currently working on an small refactor to improve the user experience.

venkataanil:
> > Please let me know what kube-burner's standards this measurement is not following so that we can improve it.
> 
> as we discussed, the measurement is not standard, and doesn't work OOTB, it needs some hardcoded tricks that can be confusing, Im currently working on an small refactor to improve the user experience.

Thanks @rsevilla87 Please add me once those PRs are ready.

PR #384, opened 2026-02-06
Add virt-density --mounts config option

jeniferh:
## Type of change

- New feature

## Description
Adding a --mounts option to control if cloud-init and emptyDisk volumes are included in the VM definition. Default is "true" which keeps the existing behavior.


jtaleric (approved):
/lgtm

Thanks @jeniferh !

PR #387, opened 2026-02-12
Obtain prometheus token when either alerting or any indexer is set

rsevilla87:
…ng are set

## Type of change

- Bug fix

## Description

When --alerting=false is passed, kube-burner doesn't collect metrics, we have to update the conditional and move this portion of the code to the main PersistentPreRun section.
 
cc @mcornea


afcollins (approved):
lgtm

PR #392, opened 2026-02-16
Add OpenSearch indexing configuration for virt scenarios

ddomingu:
## Type of change

- Bug fix

## Description

This PR adds the required configuration to enable indexing metrics and results from virt scenarios into an OpenSearch backend

- Related Issue #
- Closes #


ygalblum (changes requested):
This change breaks the current behaviour. To avoid it, local indexing should be defined by default with empty metics and alerts

ygalblum (approved):

rsevilla87 (approved):
lgtm

PR #394, opened 2026-02-16
Replace global measurements with per-job measurements in virt workloads

ddomingu:
## Type of change

- Bug fix

## Description

This PR moves measurement configuration from a global scope to a per-job level for virt workloads, allowing each job to define its own metrics independently and improving flexibility for analysis

- Related Issue #
- Closes #


ygalblum (commented):
The tests `virt-clone` `virt-ephemeral-restart` and `virt-migration` all consist of setup and actual test phases. Tracking measurements of the jobs of the setup phase is not that important. But, tracking those of the actual test are for sure required.
See my specific comments

ddomingu:
Hi, I’m closing this PR as I was looking at this from the wrong angle

PR #403, opened 2026-02-23
pprof-interval flag

rsevilla87:
## Type of change

- New feature

## Description

Add pprof-interval flag to allow configuring it.

Also, updateing the configuration to collect heap rather than CPU profiles.

- Related Issue #
- Closes #


mohit-sheth (approved):
lgtm, thanks @rsevilla87

PR #427, opened 2026-04-22
Reduce build-farm default to a more reasonable 12 iterations

afcollins:
## Type of change

- Bug fix

## Description

The default job-iterations of 100 of build-farm is more of a chaos scenario. It completely destroys a simple 3x control plane , 3x worker cluster of small 8vCPU x 18GB memory VMs.

Quick tests reveal that 14 iterations can be run without any control plane components crashing. Control plane starts to crash at only 15 iterations.

Changing the default to a more reasonable number for a better user experience out of the box.

### DIFF
--- a/pkg/workloads/build-farm.go
+++ b/pkg/workloads/build-farm.go
@@ -115,7 +115,7 @@ func NewBuildFarm(wh *workloads.WorkloadHelper) *cobra.Command {
 	}
 
 	// Standard workload flags
-	cmd.Flags().IntVar(&jobIterations, "job-iterations", 100, "Number of job iterations to create")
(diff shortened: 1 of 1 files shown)


afcollins:
Adding some pictures. The full time window shows 11-15 iterations. The annotation & cursor position show the start of 15 iterations run.

Screenshot 2026-04-21 at 17 20 06
Screenshot 2026-04-21 at 17 20 43

mcornea:
lgtm, tnx @afcollins ! Just for context for choosing this amount of iterations -  I tested the workload only on large autoscaler enabled environments targeting etcd to reach close to its 8GB max size. But I'm glad we have use cases for it on smaller envs as well.

mcornea (approved):

PR #428, opened 2026-04-23
Add cluster events to build-farm metrics profile

mcornea:
## Description

Add total cluster events to `build-farm-metrics.yml` profile


mukrishn (approved):
/lgtm

Hope that is not too bulky to index.

PR #430, opened 2026-04-24
Optimize cudn-density workload pipeline and reduce pod density

mohit-sheth:
Restructure the cudn-density workload from 8 jobs to 5 jobs:

1. Merge the separate infra job (services, NPs, egressfirewall, quotas) into the workload job. Infra objects use churn: false to protect them from churn cycles. This follows the pattern used by cluster-density-v2 and udn-density-pods.

2. Replace DaemonSet deploy/remove steps with a configurable --job-pause flag (default 1m) on the CUDN creation job, giving OVN-K time to compile the CUDN logical topology before workload deployment.

3. Reduce pod density from 9 to 4 pods per namespace:
   - Server: 6 pods -> 2 pods
   - App: 2 pods -> 1 pod
   - Client: 1 pod (unchanged) All 5 NetworkPolicies and services remain, preserving OVN-K ACL/address-set load with fewer pods to bind.

4. Remove --simple flag since infra objects are now always deployed alongside workloads.


venkataanil (approved):
LGTM

PR #437, opened 2026-05-13
Restore CPU profiling and standardize pprof duration to 60s

rsevilla87:
## Summary
- Add CPU pprof endpoints (`profile?seconds=60`) for `ovnkube-controller` and `ovnk-control-plane` alongside existing heap profiling in cluster-density-v2, node-density, node-density-cni, and node-density-heavy workloads
- Standardize CPU profiling duration from 30s to 60s in udn-density-pods, virt-udn-density, and cudn-density workloads


jtaleric (approved):
glad to see this back! Lets gooo!

PR #446, opened 2026-05-28
Remove churn boolean that is not used

afcollins:
## Type of change

- Bug fix

## Description

build-farm workload still has a boolean 'churn' flag that isn't actually used to toggle churn.

The check and logging happens in [job.go](https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/job.go#L141) and is determined from cycles and duration.

### DIFF
--- a/pkg/workloads/build-farm.go
+++ b/pkg/workloads/build-farm.go
@@ -35,7 +35,7 @@ func NewBuildFarm(wh *workloads.WorkloadHelper) *cobra.Command {
 	var qps, burst float64
 	var churnDuration, churnDelay time.Duration
 	var deletionStrategy string
-	var namespacedIterations, churn, pprof bool
+	var namespacedIterations, pprof bool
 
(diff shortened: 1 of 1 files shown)


pkg/workloads/build-farm.go
```diff
@@ -101,9 +100,6 @@ func NewBuildFarm(wh *workloads.WorkloadHelper) *cobra.Command {
 			AdditionalVars["LARGE_JOB_PERCENT"] = largeJobPercent
 
 			log.Infof("Running build-farm workload with %d job iterations across %d iterations per namespace", jobIterations, iterationsPerNamespace)
-			if churn {
```
afcollins, line 104: Logging removed happens here: https://github.com/kube-burner/kube-burner/blob/main/pkg/burner/job.go#L142-L148

rsevilla87 (approved):

PR #457, opened 2026-06-11
Adding etcd density workload

vishnuchalla:
## Type of change
- New Workload
- Documentation

## Description
Adding etcd-density from here: https://gitlab.cee.redhat.com/ancollin/etcd-density/-/tree/main?ref_type=heads which couldn't make it into wrapper in the past due to kube-burner's issue with templating the kind attribute in a k8s manifest.

Now hoping that to be fixed here: https://github.com/kube-burner/kube-burner/pull/1254

## Testing
Tested and verified in local with relevant patches.
```
../../../bin/arm64/kube-burner-ocp init -c etcd-density.yml --gc=false --qps=${QPS} --burst=${QPS} --local-indexing --log-level=debug 2>&1 | tee -a kube-burner.${ITERATIONS}its.$(date +"%F-%H%M%S").log
```

[kube-burner.10its.2026-06-11-170756.log](https://github.com/user-attachments/files/28856244/kube-burner.10its.2026-06-11-170756.log)


cmd/config/etcd-density/kubeburner-100kb.yml
afcollins, line 1: I hoped we could generate this object on the fly without storing a 100kb file into git. I think we do something similar with build-farm already?

https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/config/build-farm/build-job-configmap.yml#L10-L12

rsevilla87: correct, just use configmaps or secrets, no need to use  a custom crd

cmd/config/metrics-profiles/metrics-etcd-density.yml
afcollins, line 1: These also have some overlap with the build-farm metrics profile. Would like to consolidate the files. I think we have similar goals to capture from both workloads.

vishnuchalla: Yes, we just have one additional query here `etcdDBFragmentationBytes`. Will move to build-farm metrics profile and update here?

cmd/config/etcd-density/kubeburner-100kb.yml
vishnuchalla: Ack, on on-the-fly generation. But @afcollins are you inline with using configmaps here? I believe CRDs were kept for a purpose to ingest something completely unknown to etcd before hand.

afcollins: CRDs were used for a reason, yes. Secrets and ConfigMaps bloat other components as we scale up. CRDs only bloat etcd, apiserver. We already have configmap scale workload, it's why I created it like this.

cmd/config/metrics-profiles/metrics-etcd-density.yml
afcollins: Yep! sounds good to me

rsevilla87 (changes requested):
Hey @vishnuchalla, can you add the bats test mappings to this new workload?

https://github.com/kube-burner/kube-burner-ocp/blob/main/hack/bats_test_mappings.yml

vishnuchalla:
> Hey @vishnuchalla, can you add the bats test mappings to this new workload?
> 
> https://github.com/kube-burner/kube-burner-ocp/blob/main/hack/bats_test_mappings.yml

Done, Added.

rsevilla87 (approved):

PR #459, opened 2026-06-12
cudn-density: Add local namespace server curl to client pods

mohit-sheth:
Client pods now also curl the server in their own namespace in addition to peer namespaces within the CUDN group. The last-in-group client (previously sleep inf) now curls its local server every 10 seconds.


cmd/config/cudn-density/README.md
rsevilla87:
```suggestion
Each client generates continuous HTTP traffic to namespaces within its CUDN group, including the client's namespace:
```

venkataanil (approved):

vishnuchalla (approved):
lgtm

rsevilla87 (approved):

PR #460, opened 2026-06-15
Add gateway reachability check to cudn-density workload

venkataanil:
Add --gateway-check flag to cudn-density that deploys a gateway-checker pod in each namespace to validate north-south connectivity under CUDN scale.

## Type of change

- Refactor
- New feature
- Bug fix
- Optimization
- Documentation
- CI

- Related Issue #
- Closes #


venkataanil:
This PR is ready for the review

vishnuchalla (dismissed):
Changes look good to me.

pkg/workloads/cudn-density.go
```diff
@@ -47,6 +86,9 @@ func NewCudnDensity(wh *workloads.WorkloadHelper) *cobra.Command {
 			if iterations%namespacesPerCudn != 0 {
 				log.Fatalf("iterations (%d) must be divisible by namespaces-per-cudn (%d)", iterations, namespacesPerCudn)
 			}
+			if gatewayCheck && !l3 {
```
jtaleric: lets remove the check for l3, this solution should work with l3 or l2

cmd/config/cudn-density/deployment-gateway-checker.yml
```diff
@@ -0,0 +1,68 @@
...
+          - "-c"
+          - |
+            while true; do
+              if ping -c 1 -W 3 {{.gatewayIP}} >/dev/null 2>&1; then
+                echo "OK gateway={{.gatewayIP}}"
+              else
+                echo "FAIL gateway={{.gatewayIP}}"
+              fi
+              sleep 120
+            done
```
rsevilla87: I think this is not a correct approach, we should crash the container in case the pings fails.


```suggestion
        command:
          - "/bin/sh"
          - "-c"
          - |
            while true; do
              if ping -c 1 -W 3 {{.gatewayIP}} >/dev/null 2>&1; then
                echo "OK gateway={{.gatewayIP}}"
              else
                echo "FAIL gateway={{.gatewayIP}}"
                exit 1
              fi
              sleep 120
            done
```

Otherwise, if the gateway is not reachable, we wouldn't notice unless we grep FAIL from the logs from all these pods.

rsevilla87: wdyt @jtaleric @mohit-sheth

cmd/config/cudn-density/cudn-density.yml
```diff
@@ -199,6 +203,19 @@ jobs:
...
           namespaces_per_cudn: {{.NAMESPACES_PER_CUDN}}
 
+{{ if .GATEWAY_CHECK }}
+      - objectTemplate: np-allow-gateway-egress.yml
+        replicas: 1
+        churn: false
+        inputVars:
+          gatewayIP: {{.GATEWAY_IP}}
+
+      - objectTemplate: deployment-gateway-checker.yml
```
rsevilla87: Do we need another pod?, can't we just reuse the object that references `deployment-client.yml`?

rsevilla87: That one uses has sleep period of 10s, which I think is more appropriate than 2m

venkataanil: > Do we need another pod?, can't we just reuse the object that references `deployment-client.yml`?
I think Couple of reasons:
1. deployment-client.yml tests east-west traffic pattern. Here we are checking north-south pattern.
2. Usually there will be more stress for east-west traffic. We can increase the number of pods inside deployment-client.yml according to east-west traffic stress needed in our testing. But compared to east-west, north-south traffic will be lesser. Also we are having dedicated server pods for  deployment-client.yml. If all the deployment-client.yml stresses the external default gateway or 8.8.8.8, we may see connection drops. That will impact east-west traffic as well. It will be hard to find the reason for failures if we club both the traffic patterns in a single pod.

Customers usually have lesser namespaces and more pods. So having seperate deployments within the namespace makes sense for different traffic patterns.

I am fine to use a single pod, but I feel different deployments makes more sense. @jtaleric @mohit-sheth need your opinion as well

venkataanil: > That one uses has sleep period of 10s, which I think is more appropriate than 2m

I am following cdv2 approach.

cmd/config/cudn-density/deployment-gateway-checker.yml
jtaleric: Nice suggestion, I would also like to reduce the sleep :) to like or zero

jtaleric: I also wonder if we see situations where the ping succeeds but as we scale, it might see failures later.

cmd/config/cudn-density/cudn-density.yml
jtaleric: Might be good to have both clients doing some traffic EW and this checking the NS gw?

cmd/config/cudn-density/deployment-gateway-checker.yml
venkataanil: Does the readiness probe fail and restart the container in this case? I just added 8.8.8.8 along with cluster gateway ip. I am following cdv2 approach here. 8.8.8.8 is imilar to ROUTE_ENDPOINT and cluster gateway ip is similar to SERVICE_ENDPOINT in cdv2. I was following the cdv2 logic here 

      env:
        - name: GATEWAY_IP
          value: "{{.gatewayIP}}"
        - name: EXTERNAL_IP
          value: "8.8.8.8"
        command:
          - "/bin/sh"
          - "-c"
          - "while true; do echo ping on ${EXTERNAL_IP}; ping -c 1 -W 3 ${EXTERNAL_IP}; sleep 120; done"
        resources:
          requests:
            memory: "10Mi"
            cpu: "10m"
        imagePullPolicy: IfNotPresent
        securityContext:
          privileged: false
        readinessProbe:
          exec:
            command:
            - "/bin/sh"
            - "-c"
            - "if [ ! -f /tmp/ready.log ]; then ping -c 1 -W 3 ${GATEWAY_IP} && ping -c 1 -W 3 ${EXTERNAL_IP} && touch /tmp/ready.log; else ping -c 1 -W 3 ${GATEWAY_IP}; fi"
          periodSeconds: 10
          timeoutSeconds: 5
          failureThreshold: 3

I am fine to change if we still need to move away from cdv2 approach.

pkg/workloads/cudn-density.go
venkataanil: done

cmd/config/cudn-density/deployment-gateway-checker.yml
```diff
@@ -0,0 +1,65 @@
...
+            cpu: "10m"
+        imagePullPolicy: IfNotPresent
+        securityContext:
+          privileged: false
+        readinessProbe:
+          exec:
+            command:
+            - "/bin/sh"
+            - "-c"
+            - "if [ ! -f /tmp/ready.log ]; then ping -c 1 -W 3 ${GATEWAY_IP} && ping -c 1 -W 3 ${EXTERNAL_IP} && touch /tmp/ready.log; else ping -c 1 -W 3 ${GATEWAY_IP}; fi"
```
rsevilla87: If for some reason the traffic starts failing after a successful readinessProbe we wouldn't notice because of the /tmp/ready check. I think is better to run the commands

```suggestion
            - "ping -c 1 -W 3 ${GATEWAY_IP} && ping -c 1 -W 3 ${EXTERNAL_IP}"
```

cmd/config/cudn-density/deployment-gateway-checker.yml
```diff
@@ -0,0 +1,65 @@
...
+        image: quay.io/cloud-bulldozer/curl:latest
+        env:
+        - name: GATEWAY_IP
+          value: "{{.gatewayIP}}"
+        - name: EXTERNAL_IP
+          value: "8.8.8.8"
+        command:
+          - "/bin/sh"
+          - "-c"
+          - "while true; do echo ping on ${EXTERNAL_IP}; ping -c 1 -W 3 ${EXTERNAL_IP}; sleep 120; done"
```
rsevilla87: We already have the readinessProbe to check connectivity, just `sleep inf` here to have more control on what the pod is doing

cmd/config/cudn-density/deployment-gateway-checker.yml
venkataanil: I will drop checking 8.8.8.8

mohit-sheth (dismissed):
/lgtm
some discussion[ here ](https://redhat-internal.slack.com/archives/C0780C0J6HJ/p1781626328628659)for context

pkg/workloads/cudn-density.go
```diff
@@ -93,6 +137,7 @@ func NewCudnDensity(wh *workloads.WorkloadHelper) *cobra.Command {
 	cmd.Flags().IntVar(&iterations, "iterations", 0, "Total number of namespaces to create")
 	cmd.Flags().IntVar(&namespacesPerCudn, "namespaces-per-cudn", 5, "Number of namespaces sharing the same CUDN")
 	cmd.Flags().DurationVar(&podReadyThreshold, "pod-ready-threshold", 0, "Pod ready timeout threshold")
+	cmd.Flags().BoolVar(&gatewayCheck, "gateway-check", false, "Enable default gateway and external IP (8.8.8.8) reachability check from each namespace")
```
rsevilla87:
```suggestion
	cmd.Flags().BoolVar(&gatewayCheck, "gateway-check", false, "Enable default gateway ICMP reachability check from each namespace")
```

cmd/config/cudn-density/README.md
```diff
@@ -359,6 +372,7 @@ oc delete clusteruserdefinednetworks --all
 | [`deployment-server.yml`](deployment-server.yml) | nginx server deployment (named ports: `http`/`https`) |
 | [`deployment-app.yml`](deployment-app.yml) | sampleapp middleware deployment |
 | [`deployment-client.yml`](deployment-client.yml) | curl client with cross-namespace traffic + [readiness probes](#cross-namespace-traffic-pattern) |
+| [`deployment-gateway-checker.yml`](deployment-gateway-checker.yml) | Gateway reachability checker (pings default gateway, `--gateway-check` only) |
```
rsevilla87: Remove

rsevilla87 (approved):

PR #467, opened 2026-06-22
Bump kube-burner to v2.7.3

rsevilla87:
Bump kube-burner to v2.7.1

## Changes
- Updates kube-burner dependency from v2.7.0 to v2.7.1

## Release Notes
See upstream release notes: https://github.com/kube-burner/kube-burner/releases/tag/v2.7.1

### Key Changes in v2.7.1:
- Bug fix: Removed authorization fatal error
- Performance: Optimized deletion code using DeleteCollection
- Removed achieved QPS metric from code and bundled Grafana dashboards
- Added MicroShift support
- Added dynamic discovery for templated kinds
- Enhanced object churn with repeatEveryNIterations support

### Deprecations:
- The `beforeCleanup` parameter has been deprecated (not a breaking change, but should be migrated away from)

No breaking changes requiring immediate action.

cc: @mohit-sheth


rsevilla87:
## CI Fix Applied

Fixed the test failures caused by missing `check.sh` script. The issue was that several virt workload configs reference `./check.sh` in their hook commands, but the script was only available in the `scripts/` directory.

### Root Cause
When kube-burner v2.7.1 executes workload hooks, it looks for the script in the workload's current working directory. The embedded filesystem has the scripts at `config/scripts/check.sh`, but the hooks were looking for `./check.sh` relative to where they execute.

### Solution
Copied `check.sh` from `cmd/config/scripts/` to each virt workload directory that needs it:
- virt-capacity-benchmark
- virt-ephemeral-restart  
- virt-clone
- virt-clone-multi
- virt-parallel

This matches the existing pattern used by `virt-migration`, which already has its own copy of `check.sh`.

The CI should now pass. ✅

rsevilla87:
## Why this was working before

Great question! The issue is related to how kube-burner v2.7.1 deprecated the `beforeCleanup` parameter in favor of hooks.

### Before (v2.7.0 and earlier)
When using the deprecated `beforeCleanup` parameter:
```yaml
beforeCleanup: "./check.sh check_vm_running ..."
```

The execution flow was:
1. `RunShellCmd()` → `GetScriptsReader()` → `getEmbedReader()`
2. **Checked the embedded filesystem first** at `config/scripts/check.sh`
3. Fell back to local filesystem if not found in embedded FS
4. This meant `./check.sh` resolved to the embedded `config/scripts/check.sh`

### After (v2.7.1)
When using hooks with `when: beforeCleanup`:
```yaml
hooks:
- cmd: ["./check.sh", "check_vm_running", ...]
  when: beforeCleanup
```

The execution flow is:
1. `executeForegroundHook()` → `exec.CommandContext()` directly
2. **No embedded filesystem lookup** - just executes from current working directory
3. `./check.sh` must exist in the local directory where the workload runs

### The Fix
Since hooks don't have the embedded filesystem lookup that `beforeCleanup` had, each workload directory needs its own copy of `check.sh` - just like `virt-migration` already has.

This is actually a cleaner approach as it:
- Makes the dependency explicit (each workload has the script it needs)
- Allows per-workload customization if needed (as seen in virt-migration's enhanced version)
- Removes the "magic" filesystem resolution behavior

venkataanil (approved):
lgtm

PR #476, opened 2026-07-01
Split cudn-density infra and pod deployment into separate execution groups

kyrtapz:
Moves deployments from group 2 to group 3 so infra objects (services, network policies, egressfirewall, quotas) are fully created and settled before pods start binding ports. The existing --job-pause flag now applies to both group transitions.

This can be important at large scales, especially if OVN doesn't incrementally process objects created during infra phase.

### DIFF
--- a/cmd/config/cudn-density/cudn-density.yml
+++ b/cmd/config/cudn-density/cudn-density.yml
@@ -130,6 +130,7 @@ jobs:
         churn: false
         group:
           id: 2
+          pauseBeforeGC: {{.JOB_PAUSE}}
 
       - objectTemplate: service.yml
         replicas: 1
@@ -210,21 +211,21 @@ jobs:
       - objectTemplate: deployment-server.yml
         replicas: 1
         group:
-          id: 2
+          id: 3
         inputVars:
           podReplicas: 2
 
       - objectTemplate: deployment-app.yml
         replicas: 1
         group:
-          id: 2
+          id: 3
         inputVars:
           podReplicas: 1
 
       - objectTemplate: deployment-client.yml
         replicas: 1
         group:
-          id: 2
+          id: 3
         inputVars:
           podReplicas: 1
           namespaces_per_cudn: {{.NAMESPACES_PER_CUDN}}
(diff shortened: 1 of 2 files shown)


venkataanil:
I partially agree for this change. Earlier I wanted to have a seperate job/group (the last one) for deployment_client.yml so that server creation won't impact client podready latency. However we see mostly the customers test apps creating all the resources together and esculate about ovn-controller bottlenecks. In our standard cluster-density-v2 workload, we create all the resources together and we have a stable pod ready latency measurement for reporting. So better we keep all the resources together and only tweak during manual testing for specific purposes. @rsevilla87 @jtaleric @mohit-sheth

cmd/config/cudn-density/cudn-density.yml
```diff
@@ -130,6 +130,7 @@ jobs:
         churn: false
         group:
           id: 2
+          pauseBeforeGC: {{.JOB_PAUSE}}
```
rsevilla87, line 133: Why did you add this, isn't it redundant with jobPause?

venkataanil: Looks like Patryk wants this order with grouping
1. create namespace, cudn, wait for {{.JOB_PAUSE}} duration
2. configmap, service, service-headless, service-server, np-deny-all, wait for {{.JOB_PAUSE}} duration
3. deployment app, server, client

kyrtapz:
/hold
Not seeing any forced recomputes with the latest OVN version.

cmd/config/cudn-density/cudn-density.yml
rsevilla87: No, it won't work like that, the parameter `pauseBeforeGC` makes kube-burner to pause the execution before garbage collecting the group, this garbage collection happens after the group completes and `gc: true` at group level, which is note the case. I think we don't need this paramater

rsevilla87: Reference: https://kube-burner.github.io/kube-burner/latest/reference/configuration/?h=group#grouped-execution

kyrtapz:
No forced recomputes during group 2 + lower POD QPS make this change not worth it for now.

PR #478, opened 2026-07-01
Reduce datapoints from podStatusCount metric

rsevilla87:
## Type of change

- Optimization

## Description

When pods are not in a phase the metric has value 0, skipping those datapoints can help to reduce the number of documents generated significantly 

- Related Issue #
- Closes #


vishnuchalla (commented):
lgtm

vishnuchalla (approved):

PR #480, opened 2026-07-01
Enable ssh for virt-density

jtaleric:
## Type of change
- New feature

## Description
virt-density should have a ssh check. 

- Related Issue #
- Closes #

### DIFF
--- a/cmd/config/scripts/check.sh
+++ b/cmd/config/scripts/check.sh
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+
+# Enable persistent logging and command tracing
+LOG_FILE="${LOG_FILE:-virt-ssh-check-$(date +%Y%m%d-%H%M%S).log}"
+exec > >(tee -a "$LOG_FILE") 2>&1
+
 COMMAND=$1
 LABEL_KEY=$2
 LABEL_VALUE=$3
@@ -7,6 +12,17 @@ IDENTITY_FILE=$5
 REMOTE_USER=$6
 EXPECTED_ROOT_SIZE=$7
 EXPECTED_DATA_SIZE=$8
+NAMESPACE_LABEL="${9:-}"
+
+if [ -n "${NAMESPACE}" ] && [ -n "${NAMESPACE_LABEL}" ]; then
+    echo "Error: Cannot specify both namespace and namespace label selector. Use one or the other." >&2
+    exit 1
+fi
+
+if [ -z "${NAMESPACE}" ] && [ -z "${NAMESPACE_LABEL}" ]; then
+    echo "Error: Must specify either namespace or namespace label selector" >&2
+    exit 1
+fi
 
 # Wait up to ~60 minutes
 MAX_RETRIES=130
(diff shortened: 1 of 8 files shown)


cmd/config/virt-density/virt-density.yml
```diff
@@ -46,12 +46,44 @@ jobs:
...
     namespaceLabels:
+      kube-burner.io/test-name: virt-density
       security.openshift.io/scc.podSecurityLabelSync: false
       pod-security.kubernetes.io/enforce: privileged
       pod-security.kubernetes.io/audit: privileged
       pod-security.kubernetes.io/warn: privileged
+{{ if .SSH_CHECK }}
+    hooks:
+{{ if .NAMESPACED_ITERATIONS }}
+    - cmd: ["../scripts/check.sh", "check_vm_running", "kube-burner.io/job", "virt-density", "", "{{ .privateKey }}", "fedora", "", "", "kube-burner.io/test-name=virt-density"]
```
vishnuchalla: Just mentioning the script name should work as kube-burner explicitly looking into scripts/ folder through the file system.

capolrik (dismissed):

cmd/config/virt-density/virt-density.yml
rsevilla87: You don't need to pass `kube-burner.io/test-name`, namespaces are also labeled automatically with `kube-burner.io/job/virt-density`

rsevilla87: Same applies to the other workloads

rsevilla87 (changes requested):

vishnuchalla:
Changes looks legit for virt-density.

jtaleric:
@vishnuchalla @rsevilla87 @venkataanil Ready for review 🙏🏽

cmd/config/scripts/check.sh
```diff
@@ -1,4 +1,9 @@
 #!/usr/bin/env bash
+
+# Enable persistent logging and command tracing
+LOG_FILE="${LOG_FILE:-virt-ssh-check-$(date +%Y%m%d-%H%M%S).log}"
+exec > >(tee -a "$LOG_FILE") 2>&1
```
vishnuchalla, line 5: Do we need this log file? Looking at the failing actions: https://github.com/kube-burner/kube-burner-ocp/actions/runs/30897977761/job/91955320546
```
   time="2026-08-04 10:40:15" level=error msg="Error executing hooks for beforeCleanup: foreground hook execution failed at 'beforeCleanup': hook failed at 'beforeCleanup' after 2.326128ms: exit status 2" file="job.go:342"
```
fails with exit status 2 within 2 ms. This script seems to be dying before it get to any actual commands. I think it better to remove the block if not required. 

If the test fails even after that it worth checking this signature change.
```
  # Old: check_vm_running(vm)
  # New: check_vm_running(vm_namespace, vm_name)
```
which I think is backward compatible, less likely to be the cause for failure.

jtaleric: yes, the logfile is to determine which VM failed to ssh to and diagnose

vishnuchalla: Isn't that already being logged here: https://github.com/kube-burner/kube-burner-ocp/pull/480/changes#diff-0ad821df511047289b1de3007b74bdb97a2e8e1cac925193a179318c576f4608R157?

jtaleric: hm, I don't recall being able to debug which VM's failed so I added the debug log..

rsevilla87:
Just FYI @vishnuchalla @jtaleric , CI tests are failing probably because "sh" is being used by the GHA runner instead "bash". This PR should fix it https://github.com/kube-burner/kube-burner/pull/1291

jtaleric:
@rsevilla87 I will rebase one that merges.

vishnuchalla:
> @rsevilla87 I will rebase one that merges.

Ready for the rebase. Please update

vishnuchalla:
lgtm

PR #493, opened 2026-08-04
Add --prometheus-token-file flag for automatic token refresh

Sandeepyadav93:
Type of change

  - New feature

  Description

  The Prometheus bearer token is set once at startup and never refreshed. For providers like GKE (GMP) and
  Azure, long-running workloads (>1 hour) can see the token expire mid-run, causing 401 errors.

  This adds a --prometheus-token-file CLI flag that points to a file containing the bearer token. The token is
  re-read from the file on every Prometheus HTTP request, allowing an external process (e.g., a cron job
  running gcloud auth print-access-token) to rotate the token without restarting kube-burner-ocp.

  - Add --prometheus-token-file CLI flag alongside existing --prometheus-token
  - Wire PrometheusTokenFile through WorkloadHelper and index subcommand
  - tokenFile takes precedence over token when both are set (with a warning log)
  - Existing --prometheus-token flow is completely untouched
  - Uses upstream kube-burner v2.8.0 and go-commons v2.3.8


Sandeepyadav93:
@vishnuchalla PTAL

go.mod
```diff
@@ -5,9 +5,9 @@ go 1.25.0
 toolchain go1.25.9
 
 require (
-	github.com/cloud-bulldozer/go-commons/v2 v2.3.4
+	github.com/cloud-bulldozer/go-commons/v2 v2.3.8
 	github.com/google/uuid v1.6.0
-	github.com/kube-burner/kube-burner/v2 v2.7.3
+	github.com/kube-burner/kube-burner/v2 v2.8.0
```
vishnuchalla: Its v2.8.1 now

Sandeepyadav93: thanks, done

PR #494, opened 2026-08-04
Add --cleanup and --cleanup-only flags to virt-workload scenarios

ddomingu:
## Type of change

- New feature

## Description

Add --cleanup and --cleanup-only flags to virt-workload scenarios


pkg/workloads/virt-density.go
```diff
@@ -78,5 +94,7 @@ func NewVirtDensity(wh *workloads.WorkloadHelper) *cobra.Command {
 	cmd.Flags().DurationVar(&churnDelay, "churn-delay", 2*time.Minute, "Time to wait between each churn")
 	cmd.Flags().IntVar(&churnPercent, "churn-percent", 10, "Percentage of job iterations that kube-burner will churn each round")
 	cmd.Flags().StringVar(&churnMode, "churn-mode", string(config.ChurnObjects), "Either namespaces, to churn entire namespaces or objects, to churn individual objects")
+	cmd.Flags().BoolVar(&cleanupOnly, "cleanup-only", false, "Only cleanup the resource created by the previous run. Do not run the test.")
+	cmd.Flags().BoolVar(&cleanup, "cleanup", false, "Cleanup the resource created by the test.")
```
vishnuchalla: Terminology is a bit different with what we have today. We use `cleanup` to cleanup previous run's resources and `gc` to cleanup resources created by the current test instead. Can virt workloads inherit the same terminology?

capolrik: The `gc` option is already working with virt workloads. But I am fine to have a `cleanupOnly` or `gcOnly` option

ddomingu: I was following the same format and style used by the scenarios that already implement `--cleanup` and `--cleanup-only`: [virt-capacity-benchmark](https://github.com/kube-burner/kube-burner-ocp/blob/0d5e0bab04fec052c3e22f8e48a04fdc51f17866/pkg/workloads/virt-capacity-benchmark.go#L175-L176), [virt-clone-multi](https://github.com/kube-burner/kube-burner-ocp/blob/0d5e0bab04fec052c3e22f8e48a04fdc51f17866/pkg/workloads/virt-clone-multi.go#L141-L142) and [virt-parallel](https://github.com/kube-burner/kube-burner-ocp/blob/0d5e0bab04fec052c3e22f8e48a04fdc51f17866/pkg/workloads/virt-parallel.go#L199-L200). However, if we want to align the terminology with the other scenarios for consistency, I'm happy to update it

vishnuchalla: Yes please. Let's change the others too to be consistent with `cleanup` and `gc`.

ddomingu: Done! I also updated the documentation and aligned the remaining virt-workload scenarios that already support the `--cleanup` and `--cleanup-only` flags

vishnuchalla: Thanks for updating. Needs rebase.

pkg/workloads/virt-clone-multi.go
```diff
@@ -145,8 +145,8 @@ func NewVirtCloneMulti(wh *workloads.WorkloadHelper) *cobra.Command {
 	cmd.Flags().DurationVar(&jobIterationDelay, "job-iteration-delay", 1*time.Minute, "Delay between namespace iterations")
 	cmd.Flags().StringVarP(&testNamespaceBaseName, "namespace", "n", virtCloneMultiTestName, "Base namespace name for the test")
 	cmd.Flags().StringSliceVar(&metricsProfiles, "metrics-profile", []string{"metrics-aggregated.yml"}, "Comma separated list of metrics profiles to use")
-	cmd.Flags().BoolVar(&cleanupOnly, "cleanup-only", false, "Only cleanup the resources created by the previous run. Do not run the test.")
-	cmd.Flags().BoolVar(&cleanup, "cleanup", false, "Cleanup the resources created by the test.")
+	cmd.Flags().BoolVar(&cleanup, "cleanup", false, "Cleanup resources created by previous runs")
+	cmd.Flags().BoolVar(&gc, "gc", false, "Garbage collect resources after the current run")
```
vishnuchalla: Why do we need this again while we have the `gc` flag in ocp.go already: https://github.com/kube-burner/kube-burner-ocp/blob/main/cmd/ocp.go#L69?

ddomingu: You're right. We don't need the workload-specific `--gc` flag. Turns out most of these virt workloads weren't actually using the global `--gc` flag because their yaml configs were missing `gc: {{.GC}}` in the global section. So I've:

1. Removed the workload-specific `--gc` flags
2. Added `gc: {{.GC}}` to the yaml configs that didn't have it
3. Kept only `--cleanup` for a different use case

Now we have:
- `--cleanup`: cleans up resources from previous runs (doesn't run the test)
- `--gc`: garbage collects after the current run finishes (global flag)

vishnuchalla (commented):
lgtm

PR #502, opened 2026-08-10
Add static MAC binding workaround for cudn-density (OCPBUGS-85627)

mohit-sheth:
## Summary
- **This is a temporary patchset** — the workaround should be removed once OCPBUGS-85627 is fixed upstream
- Adds `--static-mac-binding-workaround` flag to `cudn-density` that deploys a privileged DaemonSet to inject static OVN MAC bindings on all nodes before CUDN creation
- Works around [OCPBUGS-85627](https://redhat.atlassian.net/browse/OCPBUGS-85627) where dynamic MAC bindings can cause connectivity issues in CUDN environments
- The workaround runs as a separate kube-burner job (with SA, RoleBinding, Secret, and DaemonSet) before the main cudn-density job, and is cleaned up automatically during GC
- The flag value is automatically tracked in indexed `SummaryMetadata.workloadFlags`, so runs with/without the workaround can be distinguished in dashboards
- Also adds `--deletion-strategy` flag to cudn-density (missing compared to other workloads)

## Test plan
- [x] Run `cudn-density --iterations 5 --static-mac-binding-workaround` — workaround DaemonSet created (24/24 pods ready), CUDNs created successfully (rc 0)
- [x] Run `cudn-density --iterations 5 --gc=true` without workaround flag — no workaround resources created, GC works (rc 0)
- [x] Run `cudn-density --iterations 5 --gc=true --deletion-strategy=gvr` — GVR deletion works (rc 0)
- [ ] Verify workaround flag value appears in indexed metadata for run comparison

### DIFF
--- a/cmd/config/cudn-density/cudn-density.yml
+++ b/cmd/config/cudn-density/cudn-density.yml
@@ -44,6 +44,37 @@ metricsEndpoints:
 {{ end }}
 
 jobs:
+{{ if .OCPBUGS_85627_WORKAROUND }}
+  - name: cudn-density-workaround
+    namespace: redhat-ocpbugs-85627-wa
+    jobIterations: 1
+    qps: 20
+    burst: 20
+    namespacedIterations: false
+    preCreateNamespaces: true
+    waitWhenFinished: true
+    skipIndexing: true
+    cleanup: false
+    jobPause: 1m
+    namespaceLabels:
+      security.openshift.io/scc.podSecurityLabelSync: false
+      pod-security.kubernetes.io/enforce: privileged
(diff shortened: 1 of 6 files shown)


jtaleric (commented):
great idea @mohit-sheth

jtaleric:
@mohit-sheth maybe add to the commit message, this is a temporary patchset. 

We should also store a metadata field so we can track when we have and do not have this workaround present.

mohit-sheth:
@jtaleric ack
wrt metadata workload flags are already automatically added to SummaryMetadata

venkataanil (approved):
/lgtm

PR #503, opened 2026-08-11
[PERFSCALE-5477] Enabling PPROF collection for Kubelet and CRIO components in cluster-density-v2 workload

sraviteja-maker:
## Type of change

- New feature

## Description

To collect CPU and Memory Profiling data for Kubelet and CRIO components on worker nodes while executing the cluster-density-v2 workload.

- Related Issue #
- Closes #

### DIFF
--- a/cmd/config/cluster-density-v2/cluster-density-v2.yml
+++ b/cmd/config/cluster-density-v2/cluster-density-v2.yml
@@ -15,6 +15,8 @@ global:
     - name: pprof
       pprofInterval: {{.PPROF_INTERVAL}}
       pprofDirectory: pprof-data
+      nodeAffinity:
+        node-role.kubernetes.io/worker: ""
       pprofTargets:
       - name: ovnkube-controller-heap
         namespace: "openshift-ovn-kubernetes"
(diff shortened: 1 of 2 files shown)


cmd/config/cluster-density-v2/cluster-density-v2.yml
```diff
@@ -33,6 +38,21 @@ global:
         labelSelector: {app: ovnkube-control-plane}
         url: http://localhost:29108/debug/pprof/profile?seconds=60
 {{ end }}
+{{ if .PPROF_KUBELET }}
```
vishnuchalla, line 41: For these kubelet and crio, on replica should be fine. More context here: https://github.com/kube-burner/kube-burner/pull/1285

rsevilla87: Let's bump cut a new release and bump the dependency

vishnuchalla: Yeah, sure. I need this in before: https://github.com/kube-burner/kube-burner-ocp/pull/519

rsevilla87 (commented):
This is a nice feature, however I feel that we can make it more ambitious, not only adding it to cluster-density but to other workloads too.

Also, having per-component flag, like kubele or crio does not scale well, it would be nice to allow passing a list of components, including ovn ones.i.e:

```
--pprof-components=crio,kubelet,ovnkube-controller,etcd
```

Also, to avoid repeating ourselves, I think we can think of a way to template the pprof configuration lines, so we can merge it dynamically with the workload configuration.

cc @vishnuchalla @jtaleric @afcollins

vishnuchalla:
@sraviteja-maker do you have cycles to update this PR with the optimizations suggested?

PR #510, opened 2026-08-18
Remove cleanup flags from virt workloads

rsevilla87:
## Type of change

- Bug fix
- CI

## Description

Running kube-burner with these flags is no longer required as gc is enabled by default


vishnuchalla:
lgtm

PR #517, opened 2026-08-21
Add VM cloning support to virt-parallel and virt-capacity-benchmark

ddomingu:
## Type of change

- New feature

## Description

Enable VM creation from base images instead of direct container disk pulls. New flags allow testing storage cloning performance and reducing registry load:
  - `--use-base-image`: Clone VMs from a base `DataVolume`
  - `--use-snapshot`: Clone from `VolumeSnapshot` or direct PVC clone

The workflow creates a base `DataVolume` from the container image, optionally creates a `VolumeSnapshot`, then clones all VMs from the snapshot or PVC.


rsevilla87 (approved):
lgtm

PR #518, opened 2026-08-26
Fix dv-clone cleanup to clear VolumeSnapshot finalizers

amitosw15:
## Type of change

- Bug fix
- CI

## Description

`dv-clone` defaults to `--use-snapshot=true`, which creates VolumeSnapshots. Tearing down namespaces with a plain `oc delete ns` leaves finalizers on non-`Delete` VolumeSnapshotContent, so the namespace sticks in `Terminating` and reruns collide.

Add a `--cleanup` flag (mirroring `virt-ephemeral-restart` and other virt workloads) that uses the shared `cleanupTestNamespaces` helper to clear snapshot finalizers and orphaned PVs. Update the bats test to call `dv-clone --cleanup` instead of `oc delete ns`.

## Test plan

- [x] `go build ./...` and `go vet ./pkg/workloads/...`
- [x] `dv-clone --cleanup` verified on live cluster (3×, exit 0, no hang)
- [ ] CI bats `workload:dv-clone` (requires `ok-to-test` label)

Made with [Cursor](https://cursor.com)


vishnuchalla (approved):
/lgtm

